Issue #6, opened 2015-11-07
New launcher icon wanted


ViliusKraujutis:
Now legacy icon looks like this:
![apps_icon](https://cloud.githubusercontent.com/assets/1859636/11012024/f1daee46-84f6-11e5-9131-81f9c9a6c017.png)

It could be updated to to match style of other [Vilnius apps](https://play.google.com/store/apps/developer?id=Vilnius) 
![image](https://cloud.githubusercontent.com/assets/1859636/11012039/27482224-84f7-11e5-8be2-9bd9e6d3e615.png)

Or maybe include Vilnius Logo 
![image](https://cloud.githubusercontent.com/assets/1859636/11012051/595434ba-84f7-11e5-8e0d-72e0b0eee0cc.png)

ViliusKraujutis:
Google's icon guidelines are here: https://www.google.com/design/spec/style/icons.html#

PR #7, opened 2015-11-07
Replace stub launcher icon with legacy icon

ViliusKraujutis:
But also raising a new issue #6, because I guess icon could also be refreshed.


ViliusKraujutis:
10 days passed, no objections and it's super simple change.

Issue #2, opened 2015-10-29
Package name


ViliusKraujutis:
I suggest we make package name `lt.vilnius.tvarkau` instead of `uk.co.es4b.tvarkau_vilniu`.

I dislike when small workshops adds their domain in apps they make. Look how it looks in Google Play Store URL:
`https://play.google.com/store/apps/details?id=uk.co.es4b.tvarkau_vilniu`

Instead it could look like this:
`https://play.google.com/store/apps/details?id=lt.vilnius.tvarkau`.

Cons:
All current users will have to reinstall new application.

tadaskay:
Once we launch, we can release an update to the old app with some sort of redirection. Thus it wouldn't be a problem.

Regarding the package, I honestly think your proposal fits better than uk.*

ViliusKraujutis:
And as agreed on Slack [here](https://codeforvilnius.slack.com/archives/tvarkau-vilniu/p1446228463000066) too, I'm closing this issue now.

ViliusKraujutis:
Some reasons against changing the name https://codeforvilnius.slack.com/archives/tvarkau-vilniu/p1447689548000152

PR #8, opened 2015-11-15
Added a launch activity to the project.

overengineered:
The start activity far from final as the design is not available yet. But it will give us a starting point for working on other activities, for which we do have designs.


.gitignore
```diff
@@ -1,8 +1,7 @@
 *.iml
 .gradle
-/local.properties
-/.idea/workspace.xml
```
ViliusKraujutis, line 4: Did you remove those intentionally?
According to [JetBrains](https://intellij-support.jetbrains.com/hc/en-us/articles/206827587-How-to-manage-projects-under-Version-Control-Systems) this line should be kept.

.gitignore
```diff
@@ -1,8 +1,7 @@
...
-/.idea/libraries
+.idea
 .DS_Store
+/local.properties
```
ViliusKraujutis, line 5: Was it different sorting?

.gitignore
```diff
@@ -1,8 +1,7 @@
 *.iml
 .gradle
-/local.properties
-/.idea/workspace.xml
-/.idea/libraries
+.idea
```
ViliusKraujutis, line 3: Ah... Looks like you removed `.idea` files intentionally. Right?

app/src/main/AndroidManifest.xml
```diff
@@ -5,9 +5,23 @@
...
-        android:supportsRtl="true"
         android:theme="@style/AppTheme">
 
+        <activity android:name="lt.vilnius.tvarkau.StartActivity" />
+
+        <!-- This strangely named alias is here for backwards compatibility.
+             Renaming activity with LAUNCHER category means that shortcuts created by users
+             on their phones with previous version of the app will disappear or stop working. -->
+        <activity-alias
+            android:name="apps.factory.emiestas.MainActivity"
```
ViliusKraujutis, line 16: Do we really need it? [This](http://developer.android.com/guide/topics/manifest/activity-alias-element.html) says `The target must be in the same application as the alias and it must be declared before the alias in the manifest`. If we change app's package name, then it means we'll have different application and those previously created links would not work. Isn't this the case?

ViliusKraujutis:
Everything else LGTM :+1:

.gitignore
overengineered, line 3: Yes, I removed .idea files because Android Studio generates them when opening project, and creates them differently on different systems. And I did some sorting.

app/src/main/AndroidManifest.xml
overengineered, line 16: Well, I'm working under assumption that we'll release the new app with the old package id. Otherwise we'll be left managing 2 apps instead of 1.

ViliusKraujutis, line 16: Then, please reopen https://github.com/vilnius/tvarkau-vilniu/issues/2 and let's discuss that with your pros/cons for that.
I would rather manage 2 apps instead of 1, than have that ugly package id. Then we force those few users to update to new app and unpublish this old app too.

If we don't change this legacy package id now, then we won't change it in the future for sure.

RenatBuga:
Vilnius, how this issue is related to the problem we have set to solve with this project? 

1. Complicated registration
2. Bad UX

May be we just focus on what's important and leave asside the things that are unrelated to our goals? Each issue like this is a barrier in our way towards our goals.

We are very limited in our time and we spend a lot of it in discussing minor things instead of doing major things.

RenatBuga:
Vilius, how this issue is related to the problem we have set to solve with this project? 
1. Complicated registration
2. Bad UX

May be we just focus on what's important and leave aside the things that are unrelated to our goals? Each issue like this is a barrier in our way towards our goals.

We are very limited in our time and we spend a lot of it in discussing minor things instead of doing major things.

app/src/main/AndroidManifest.xml
ViliusKraujutis, line 16: BTW, this is not a blocker for merge. Just a minor issue to discuss while the process is moving on. [Here](https://github.com/vilnius/tvarkau-vilniu/pull/8#issuecomment-157017925) @RenatBuga probably wanted to reply in this thread. I agree with focus. But this change is better now or later will be worse (and probably not worth it, so - now or never).

RenatBuga, line 16: As I mentioned, our limited resource is our time. 
Thinking/Discussing/Typing - all this takes our precious time that we can use instead for droping few extra lines of code.

I vote FOR changing the package name as long as doesn't put under risk to loose any of 786 existing users. However, I'm encouraging everyone to keep focusing on our goals and use our time with the greatest benefit to the project.

.gitignore
tadaskay, line 4: I saw that the whole `.idea` folder is now git-ignored, which is the right thing.

ViliusKraujutis:
Since we are 3 android developers and I :+1:'ed already, so it's majority (when counting Juozas in)! Merging, so we could move on with #9, but @GediminasZukas, please take a look to this PR.

RenatBuga:
The voting results: 4 against 1 to keep the package name as it is.

PR #11, opened 2015-11-17
Formated the look of List of Problems

RenatBuga:
Done
- Adjusted the look of the list of problems as per design
- Added placeholder for pics

Needs to be fixed:
- The background has to be white color
- The status label look has to be changed from just being colored to label_report_status_1.xml (blue) or label_report_status_2.xml (green)


vycius:
Screenshots look awesome. Few suggestions from my side:
- I think we can throw away login screen and add button `Sign in with Google` to home screen. If user doesn't click `Sign in with Google` button it means he is a guest.
- I don't like our home screen. User should see problems list/map as soon as he opens our app. IMHO Navigation drawer and list/map should look more user friendly. We should discuss this option. 
  materialnavigationdrawer
- For `Add report` screen I would recommend to add `send` button to actionbar. If user enters a lot of text to problem description buttons  will be hidden. 
- How user will select location of problem ? 
- Most of the locations will have only coordinates. Showing address is ok, but showing coordinates is not user friendly. Can we show small map with Map marker instead ?

tadaskay:
How about we merge this as-is and then make improvements separately? This has been sitting around for 20+ days without any feedback so I assume we're more or less on the same page.

vycius:
:+1: Let's merge this.

ViliusKraujutis:
:+1: ok. Adding resources... will require cleanup afterwards if some of it would not be used.  Merging.

Creatium:
@vycius We have made a few other concepts which you can find in our Slack discussion.

PR #14, opened 2015-12-11
Fix travis build

vycius:
Currently we don't have any tests so there is no need to run `connectedCheck` in travis. Leaving check if project builds. I will add running tests on device with first `tvarkau-vilniu` app tests.  Also fixed travis icon link.


ViliusKraujutis:
:+1:

PR #19, opened 2016-01-21
Home screen + my profile

vycius:
After discussion we agreed that navigation drawer is not best option. Getting back to home screen with some additional changes. Layouts can be improved, but it's not in the scope of this Pull request. 

@ViliusKraujutis give thumbs up and comments

screenshot_20160121-103730
screenshot_20160121-103757
screenshot_20160121-115920


narkusv:
Profile layout definitely could use some polishing, otherwise, imho, :+1:

ViliusKraujutis:
:+1:

PR #20, opened 2016-01-22
Google+ login implemented. 

narkusv:
Title talks for it self. User data is not being fetch from google+ when i goes into profile screen. If user has decided to browse anonymously then he his transferred back to SignInActivity. 

All classes that extend from BaseActivity is able to get Google+ user data. 

Also minor changes in NewProblemActivity layout - Toolbar introduced instead of action bar, and send button transferred there. 

First pull request ever - http://devopsreactions.tumblr.com/post/137275323038/reviewing-the-interns-commits


vycius:
Travis build has failed. Please check that.

app/src/main/java/lt/vilnius/tvarkau/SignInActivity.java
```diff
@@ -0,0 +1,104 @@
...
+import butterknife.OnClick;
+
+
+/*
+    Activity which handles log in of user.
+    It does not matter whether user decides to do it,
+    he will be transferred to MainActivity
+
+ */
+public class SignInActivity extends AppCompatActivity implements GoogleApiClient.OnConnectionFailedListener{
```
vycius: Can we extend `SignInActivity` with `BaseActivity` ? In such case we won't need to duplicate logic of google login.

narkusv: Not sure. Had been thinking about that, but decided to split them up, as SignInActivity is purely just for authentication while baseactivity is just signing in with known details. But ATM i think it makes sense to extend from BaseActivity

vycius: I have been been thinking do we really need separate activity for login ? In home screen we have sign in button. After clicking this button user will get login dialog with account selection. I think having separate `SignInActivity` is unnecessary and also not very very convenient for user (he needs one more click to just select if he wants to sign in or not). What do you think about it ?
screenshot_20160123-112633

RenatBuga: I propose to remove the login button from home screen and show it up in the following cases:
1. The user doesn't have a profile yet and clicks My Profile on home screen for the first time
2. The user doesn't have a profile yet and selects "Contact me in case of need" in Nauja problema screen.

my_profile

vycius:
One small scenario to fix:
Click My Profile -> SignInWithGoogle -> Back button -> Back button. You will end up in empty activity
screenshot_20160123-173159

app/src/main/AndroidManifest.xml
```diff
@@ -52,6 +54,12 @@
         </activity>
 
         <activity
+            android:name=".SignInActivity"
+            android:theme="@style/AppTheme"
+            />
```
vycius: In V1 we won't support horizontal layouts. I would recommend to add these two lines here:

``` xml
android:configChanges="orientation|keyboardHidden"
android:screenOrientation="portrait"
```

app/src/main/res/layout/login.xml
```diff
@@ -15,7 +15,11 @@
         android:paddingBottom="112dp"
         />
 
+
+
+
     <TextView
+        android:id="@+id/signIn"
```
vycius: Let's keep same ids style. Most ids are with `_`, not camelCased.

app/src/main/AndroidManifest.xml
narkusv: Thanks. Fixed it. About that buggy scenario - could I simply flush backstack with Intent flags? I think there is no point to let user go back to SignInActivity

vycius:
:+1: But let's wait for @ViliusKraujutis review

app/src/main/java/lt/vilnius/tvarkau/BaseActivity.java
```diff
@@ -1,19 +1,55 @@
...
+
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         Icepick.restoreInstanceState(this, savedInstanceState);
+
+
+        //Configure google sign in to request user email. Is ID and photo is included in Default Sign In.
+        GoogleSignInOptions gso = new GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
```
ViliusKraujutis: Let's keep all Sign-in stuff in a separate activity, not BaseActivity.

Or are there a reason why each activity should be able to use g+ sign-in?

app/src/main/java/lt/vilnius/tvarkau/BaseActivity.java
```diff
@@ -21,4 +57,89 @@ public void onSaveInstanceState(Bundle outState) {
...
+
+    protected void showProgressDialog(){
+        if(mProgressDialog != null){
+            mProgressDialog = new ProgressDialog(this);
+            mProgressDialog.setMessage(getString(R.string.progress_dialog_loading));
+            mProgressDialog.setIndeterminate(true);
+        }
+    }
+    protected void hideProgressDialog() {
+        if(mProgressDialog !=null && mProgressDialog.isShowing()){
```
ViliusKraujutis: Code formatting is incorrect here. Not a big deal, but... Please set up automatic code formatting to be run on each "Save" command. Otherwise, it will be formatted by someone else and "git blame" history will be removed :/

app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
```diff
@@ -110,4 +152,8 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         }
     }
 
+    private void showToast(String text){
```
ViliusKraujutis: Well, let's better inline this method instead. It's used only once. If you'd put it where it's called, then there would be no need to check, what this method does.

app/src/main/java/lt/vilnius/tvarkau/SignInActivity.java
ViliusKraujutis: Well, seems like we have code duplication here. Not good. Maybe let's extract everything g+ signin related to another class and make composition instead of code duplication?

app/src/main/java/lt/vilnius/tvarkau/fragments/MyProfileFragment.java
```diff
@@ -30,12 +35,37 @@ public static MyProfileFragment getInstance() {
...
+        setUpUserProfile();
+    }
+
+    private void setUpUserProfile(){
+        Profile profile;
+        BaseActivity baseActivity = (BaseActivity) getActivity();
+        if(baseActivity!=null){
+            profile = new Profile(baseActivity.mGoogleSignInAccount);
+
+            mProfileName.setHint(profile.getName());
```
ViliusKraujutis: Do you really want to `setHint` or `setText`?

app/src/main/java/lt/vilnius/tvarkau/BaseActivity.java
ViliusKraujutis: And let's sharedprefs store signed in user's Profile information.

app/src/main/res/values/strings.xml
```diff
@@ -16,10 +16,16 @@
...
     <string name="problem_search">Problemos paieška</string>
+    <string name="progress_dialog_loading">Kraunama</string>
 
     <string name="profile_hint_email">E-mail</string>
     <string name="profile_hint_name">Vardas</string>
     <string name="profile_telephone">Telefonas</string>
+    <string name="message_notsigned">Norint pasiekti profilio informacija prašome prisijungti</string>
+    <string name="title_notsigned">Vartotojas neprisijungęs</string>
+    <string name="problem_send">Pranešti apie problemą</string>
+    <string name="login_sign_in">Sign in with Google</string>
```
ViliusKraujutis: Hm... Let's keep language consistency in this file. We can add `/values-en/strings.xml` with all english translations.

@RenatBuga should we keep app's interface all lithuanian. Right?

ViliusKraujutis:
I'd not hurry to merge this yet. Let's discuss why we need g+ stuff in BaseActivity and code duplication in `SignInActivity`.

app/src/main/res/values/strings.xml
RenatBuga: @ViliusKraujutis for the next release - yes. The further releases need to have English as well for foreigners living in Vilnius

narkusv:
Hi, I have made additional commit with fixes discussed with @ViliusKraujutis . I Moved all Sign in stuff to helper class, although it has to use actual activity for `.enableAutoManage(activity, this)` method, which is worrying me, not sure if it is good idea to pass whole activity. **See edit**

Also i removed all g+ stuff from `BaseActivity` and all other classes except from `SignInActivity`, as we have decided to store user information locally, so there is no point at all right now to make any connection to Google.
Whole `Profile` object is stored in shared preferences with help of Gson.

Also made a macro for optimizing imports and formatting code on save. Thx for that one :)

EDIT: Accidentaly I bumped into something called `WeakReference`. Maybe referencing activity to `GoogleSignInHelper` by using it  would be just okay, and it would prevent memory leaks?

ViliusKraujutis:
Much better now. Well done, @narkusv! No show stoppers :+1: @vycius, what do you think?

app/src/main/java/lt/vilnius/tvarkau/BaseActivity.java
```diff
@@ -8,12 +8,20 @@
 /**
  * Created by Karolis Vycius on 2016-01-15.
  */
+
+/*
+    This class also is responsible for keeping user google account signed in. It is very important, that all activities, that
+    will be using google sign in, would extend this class.
+
```
vycius, line 12: Remove misleading comment.

app/src/main/java/lt/vilnius/tvarkau/entity/Profile.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "Profile{" +
+                "name='" + name + '\'' +
+                ", email='" + email + '\'' +
+                ", mobilePhone='" + mobilePhone + '\'' +
+                ", pictureUrl='" + pictureUrl + '\'' +
+                '}';
```
vycius: Just return JSON here by calling `createJsonData`

app/src/main/java/lt/vilnius/tvarkau/entity/Profile.java
```diff
@@ -0,0 +1,88 @@
...
+        this.pictureUrl = pictureUrl;
+    }
+
+    public static Profile returnProfile(Context con) {
+        return SharedPrefsManager.instance(con).getUserProfile();
+    }
+
+    public String createJsonData(Profile profile) {
+        Gson gson = new Gson();
+        return gson.toJson(profile);
```
vycius: Agree with Vilius, that this method should be changed to:

``` java
return gson.toJson(this);
```

Or you can make it `static`. I would choose first option.

app/src/main/java/lt/vilnius/tvarkau/fragments/MyProfileFragment.java
```diff
@@ -30,12 +34,37 @@ public static MyProfileFragment getInstance() {
...
+    @Override
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+        setUpUserProfile();
+    }
+
+    private void setUpUserProfile(){
+        Profile profile;
+
+        if(getActivity()!=null){
```
vycius: Formatting issue related to spaces.

``` java
if(getActivity( )!= null) {
```

 Also can you give a scenario when this if will fail ?

app/src/main/java/lt/vilnius/tvarkau/utils/GoogleSignInHelper.java
```diff
@@ -0,0 +1,68 @@
...
+
+public class GoogleSignInHelper implements GoogleApiClient.OnConnectionFailedListener {
+    public static final int RC_SIGN_IN = 9001;
+
+    private GoogleApiClient googleApiClient;
+
+    private FragmentActivity activity;
+
+
+    public GoogleSignInHelper(FragmentActivity activity) {
```
vycius: I don't like passing the whole `FragmentActivity` here. I think it would make sense to create separate `interface` with logged in callbacks  (or use callbacks from `GoogleApiClient`) and pass it to GoogleSignHelper. It's really error prone here, because `onActivityResult` of `FragmentActivity` can be easily messed up.

app/src/main/java/lt/vilnius/tvarkau/utils/SharedPrefsManager.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static final String PREF_USER_PROFILE = "UserProfile";
+    private static final String PREF_DEFAULT_USER_PROFILE = "[]";
+
+    private static SharedPrefsManager mSingleton = new SharedPrefsManager();
+    private static Context mContext;
+
+
+    public static SharedPrefsManager instance(Context context) {
+        mContext = context;
```
vycius: We need `context` only for `SharedPreferences` initiation. `getPrefs` code can be moved to `instance`. Also `private static Context mContext;` should be changed to `private static SharedPreferences mSharedPrefences`.

Furthermore I would use `ApplicationContext` for initiating `SharedPreferences`.

app/src/main/java/lt/vilnius/tvarkau/utils/SharedPrefsManager.java
```diff
@@ -0,0 +1,77 @@
...
+ * but later on it also will help us to store actual user information.
+ */
+public class SharedPrefsManager {
+    private static final String PREFS_NAME = "TVARKAU-VILNIU_PREFS";
+
+    private static final String PREF_ISANONYMOUS = "isUserAnonymous";
+    private static final Boolean DEFAULT_PREF_ISANONYMOUS = true;
+
+    private static final String PREF_USER_PROFILE = "UserProfile";
+    private static final String PREF_DEFAULT_USER_PROFILE = "[]";
```
vycius: I think what default user profile is not JSON array. Possibly you mean `{}` ?

app/src/main/java/lt/vilnius/tvarkau/utils/SharedPrefsManager.java
```diff
@@ -0,0 +1,77 @@
...
+        String json = profile.createJsonData(profile);
+        edit.putString(PREF_USER_PROFILE, json);
+        edit.commit();
+    }
+
+
+    public Profile getUserProfile() {
+        try {
+            Gson gson = new Gson();
+            String json = getPrefs().getString(PREF_USER_PROFILE, PREF_DEFAULT_USER_PROFILE);
```
vycius: I would return null if we don't have user profile. Otherwise if it's not null I would try to deserialize it.

app/src/main/java/lt/vilnius/tvarkau/utils/GoogleSignInHelper.java
```diff
@@ -0,0 +1,68 @@
...
+        activity.startActivityForResult(signInIntent, RC_SIGN_IN);
+    }
+
+
+    public void saveUserDetailsInPrefs(GoogleSignInResult result) {
+        if (result.isSuccess()) {
+            SharedPrefsManager.instance(activity).setUserAnonymous(false);
+            SharedPrefsManager.instance(activity).saveUserDetails(new Profile(result.getSignInAccount()));
+            Intent intent = new Intent(activity, MainActivity.class);
+            //Clearing backstack after login. User does not need to go back to SignInActivity.
```
vycius: Small formatting issue. One space after `//` such as 

``` java
// some comment
```

not 

``` java
//some comment
```

vycius:
:+1: I like it. Added few small comments and suggestions.

app/src/main/java/lt/vilnius/tvarkau/utils/SharedPrefsManager.java
```diff
@@ -0,0 +1,77 @@
...
+        SharedPreferences.Editor edit = getPrefs().edit();
+        edit.putBoolean(PREF_ISANONYMOUS, isAnonymous);
+        edit.apply();
+    }
+
+    public void saveUserDetails(Profile profile) {
+        SharedPreferences.Editor edit = getPrefs().edit();
+        String json = profile.createJsonData(profile);
+        edit.putString(PREF_USER_PROFILE, json);
+        edit.commit();
```
vycius: `setUserAnonymous` uses `edit.apply()`, `saveUserDetails` uses `edit.commit()`. Use same method for writing changes: `commit` or `apply` in both of these. If we don't care about return value just use `apply`.

tadaskay:
One small question: how do we plan to account for expiration and revocation of google access tokens?

app/src/main/java/lt/vilnius/tvarkau/utils/GoogleSignInHelper.java
ViliusKraujutis: mhm. Really small, that I guess it's even not covered by autoformatting. :)

app/src/main/java/lt/vilnius/tvarkau/MainActivity.java
```diff
@@ -16,6 +29,14 @@ protected void onCreate(Bundle savedInstanceState) {
         setContentView(R.layout.home);
 
         ButterKnife.bind(this);
+
+        handleSignInButton();
+    }
+
+    private void handleSignInButton() {
+        if(!SharedPrefsManager.instance(this).getIsUserAnonymous()){
```
ViliusKraujutis: Good point noticed by @vycius - use application context instead of this activity.

app/src/main/java/lt/vilnius/tvarkau/utils/SharedPrefsManager.java
ViliusKraujutis: Good points @vycius :+1:

app/src/main/java/lt/vilnius/tvarkau/utils/SharedPrefsManager.java
```diff
@@ -0,0 +1,77 @@
...
+    private static void init(Context context) {
+        mContext = context;
+    }
+
+    public SharedPreferences getPrefs() {
+        return mContext.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE);
+    }
+
+    public Boolean getIsUserAnonymous() {
+        return getPrefs().getBoolean(PREF_ISANONYMOUS, DEFAULT_PREF_ISANONYMOUS);
```
ViliusKraujutis: Change it to `return null == getPrefs().getString(PREF_USER_PROFILE, null)`. Also remove both `PREF_ISANONYMOUS` and `DEFAULT_PREF_ISANONYMOUS`.

app/src/main/java/lt/vilnius/tvarkau/utils/SharedPrefsManager.java
ViliusKraujutis: Agree, use `null` instead of `PREF_DEFAULT_USER_PROFILE`. And mark, that this method is `@Nullable`.

vycius:
:+1: @narkusv What do you think about merging this now ?

narkusv:
I hope it's all good!

PR #24, opened 2016-01-23
Remove problem detail comment button.

vycius:
Commenting on issue is not supported by municipality system. After discussion we decided to remove this button. 
@narkusv @RenatBuga 
screenshot_20160123-145630


narkusv:
:+1:

PR #27, opened 2016-01-31
Map activity rework

vycius:
Task https://trello.com/c/1Fpz6uUf/38-map-activity

screenshot_20160131-115803
screenshot_20160131-115812

@RenatBuga @narkusv


app/src/main/java/lt/vilnius/tvarkau/fragments/ProblemsMapFragment.java
```diff
@@ -1,14 +1,155 @@
...
+
+        getMapAsync(this);
+    }
+
+    private void setMarkerResources() {
+        inProgressMarker = BitmapDescriptorFactory.fromResource(R.drawable.ic_pin_drop_blue);
+        doneMarker = BitmapDescriptorFactory.fromResource(R.drawable.ic_pin_drop_green);
+    }
+
+    protected void requestGPSPermission() {
```
narkusv, line 68: I am also requesting GPS permission in NewProblem activity. Maybe we should move `requestXPermission` methods to PermissionUtils?

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemsListAdapter.java
```diff
@@ -52,25 +52,21 @@ public void onBindViewHolder(final ViewHolder holder, int position) {
...
 
+        String thumbUrl = item.getThumbUrl();
+
+        if (thumbUrl == null) {
+            holder.thumbView.setImageResource(R.drawable.ic_placeholder_list_of_reports);
+        } else {
+            Picasso.with(context).load(thumbUrl).into(holder.thumbView);
+        }
+
         holder.content.setOnClickListener(new View.OnClickListener() {
```
narkusv, line 64: IMHO it is better to create onItemClicked interface in adapter, set it in `Activity` or `Fragment`, and handle onClick methods from there.

app/src/main/java/lt/vilnius/tvarkau/fragments/ProblemsMapFragment.java
vycius, line 68: We already have method for requesting permission. This places can be changed to just:

``` java
        if(PermissionUtils.verifyAndRequestPermissions(getActivity(), GPS_PERMISSION_REQUEST_CODE,
                Manifest.permission.ACCESS_FINE_LOCATION)) {
            googleMap.setMyLocationEnabled(true);
        }
```

Although if we do this Android studio will give us a warning:
screenshot from 2016-02-01 13-36-02
I don't know what is the best way to avoid it.

vycius, line 68: Modified `PermissionUtils` methods. Now it has only ability to check only if all required permissions are granted. Remove `try/catch` from `googleMap.setMyLocationEnabled(true);`. It shows lint warnings, but IMHO it's OK, because there is check inside `PermissionUtils`.

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemsListAdapter.java
vycius, line 64: RecyclerView don't have 'setOnItemClickedListener' method. It's not really hard to create that. Do you really think it's worth creating ?

ViliusKraujutis:
:+1: lgtm

Issue #18, opened 2016-01-20
NewProblemActivity selecting photo opens camera intent directly.


narkusv:
User must be left with choice of selecting between launching camera intent and launching gallery as current user location does not necessary represent problem location.

vycius:
Also while solving this issue we might need to think about taking/picking few photos.

PR #34, opened 2016-04-29
Report problem rework

vycius:
https://trello.com/c/aFEVCk4v/35-rework-new-problem-activity
https://github.com/vilnius/tvarkau-vilniu/issues/18
@GediminasZukas @eronisko @RenatBuga @ViliusKraujutis @narkusv 
screenshot_20160429-204835
screenshot_20160429-204335
screenshot_20160429-215100


GediminasZukas:
Great work @vycius , just one issue. While testing your pull request code on my Nexus 5 device, I have found out that photo loading process is not smooth when I try to swipe right/left on the image pager area.

Let's say I pick 3 photos from gallery. Then I swipe 2 times to the left and get such empty white view for a second:
device-2016-05-02-171334
And after that sec. I see my photo:
device-2016-05-02-171407

logcat prints such messages: "Skipped 31 frames!  The application may be doing too much work on its main thread."

I think we must implement some sort of image caching and compression mechanism.

vycius:
@GediminasZukas Thanks for pointing out issue related to image loading. I am thinking what is the best way to solve it. Empty view problem can be easily fixed by using placeholder image for example image with loading indicator. Additionally we can set `offScreenPageLimit` to `ViewPager`, for instance:

``` java
            mProblemImagesViewPager.setOffscreenPageLimit(3);
```

Although setting too high number might lead to longer GC pauses. I would also consider switching to `Glide` instead of `Picasso`. According to https://inthecheesefactory.com/blog/get-to-know-glide-recommended-by-google/en Glide uses less memory and loads images faster. Do you have any more suggestions to solve this issue ?

vycius:
Changed to Glide and added off screen page limit. It works really fast now. @GediminasZukas Can you test this on your device ?

GediminasZukas:
OK, going to test now. Also, I was writing a comment just before your changes: 

I think we can set placeholder image and your suggested `offScreenPageLimit` value (default seems to be 1) no matter of this problem.

```
mProblemImagesViewPager.setAdapter(new ProblemImagesPagerAdapter<Uri>(this, photoUris) {
             @Override
             public void loadImage(Uri imageURI, Context context, ImageView imageView) {
                 Picasso.with(context).load(new File(imageURI.getPath())).into(imageView);
             }
         });
```

What if we would pass not uri's but array of bitmaps to the adapter? I think that a lot of work is going on "under the hood" in our current solution. Especially if photo size is 1 MB+. 

Would be good to know more about `Glide` from people who actually used it. Personally I never worked with this library.

GediminasZukas:
OK, tested. Just picked 3 photos from gallery and I can't see any of them (only 3 indicator circles):
device-2016-05-02-191155

vycius:
I messed up something while applying last two commits. I will take a look at it later.

vycius:
The bug was introduced by:

``` java
load(new File(imageURI.getPath()))
```

changing to 

``` java
load(imageURI)
```

It seems that nor Picasso nor Glide can't handle uris without creating File object. Reverted this line change and it should be working know.

vycius:
Regarding Bitmaps array idea it would definitely solve long loading problem between swipes. Although it works great only if you have a really small number of photos. Otherwise you will need to wait a long time to see first photo.
Current loading process with default Picasso or Glide settings:
1. Fetch photo from storage and convert it to Bitmap. 
2. Glide would resize the photo before storing it to cache. Picasso won't do that.
3. Store it to the cache. 
4. Picasso will resize it before loading to ImageView. Glide will load already resized image.
5. Load Bitmap to imageview. 
In case you need to show same photo again it has really high probability of showing photo from cache. 

Bitmaps array scenario:
1. Fetch all photos from storage and convert them to Bitmaps. This step should definitely include photo resizing or otherwise `outOfMemory` exceptions
2. Create array of Bitmaps and use them in ImageViews. 

The main Bitmap array drawback is long waiting time to load + resize all photos. Also I am a little bit afraid of memory problems http://developer.android.com/training/displaying-bitmaps/load-bitmap.html 

If recent changes increased swie-able photos performance I would be in favor of old solution. @GediminasZukas waiting for you opinion.

GediminasZukas:
I think we have to take into account functionality which will be implemented in the future - all those images with problem description will be uploaded to the backend once we are going to hit "submit" button. During this step I think it is necessary to compress or resize (thus convert images to bitmaps) the images. So maybe we could do this work just after picking the images from camera/gallery and use resulting bitmaps to solve our current problem also.

And about possible performance problem with "big" amount of photos - I think we could implement some sort of "lazy loading" mechanism to increase UX instead of blocking the UI with loading dialog or smth like that.

vycius:
For uploading photos to server you don't need to create Bitmaps from URIS (in case you don't resize them). Images are already compressed using JPEG, PNG or similar format and they should be about 1 - 1.5 MB each. Using Retrofit2 it will be really simple to send image to server without turning them into Bitmaps:

``` java
    File file = FileUtils.getFile(this, fileUri);
    RequestBody requestFile =
            RequestBody.create(MediaType.parse("multipart/form-data"), file);
```

I really like lazy loading idea. Although making it reality would require a huge amount of work.
Also in current PR we already have a similar concept to lazy loading bitmaps. We load 2 images to the right/left from uris. Additionally, let's say we have 6 photos and you swipe from first to the sixth. In case you swipe back to the first one: first picture will be loaded from LRU cache, not from storage.

GediminasZukas:
Well, for example I just checked photos taken with the camera of my old nexus 5 phone. Some .jpg format photos sizes exceeds 5 MB. So let's say user picks up such 3 photos and hits "report" button. In such situation he will loose 15 MB+ of bandwidth just for reporting the problem. Furthermore, many issues can occur related to the user mobile internet speed... So I think it's necessary to compress images to the KB zone. I believe that exists plenty image compression algorithms which can compress images without significant quality loss.

All in all, I think we could merge current PR and come back to this issues when we will be working with server side integration.

vycius:
We will investigate resizing, compression later. Added task for that in https://trello.com/c/iDvb782I/47-report-problem-api-integration

PR #37, opened 2016-05-04
Click address go to map (DELETE, not merged)

GediminasZukas:
I made some git mistakes so I created a new pull request with the same features: https://github.com/vilnius/tvarkau-vilniu/pull/40


vycius:
👍 Looks good. 
Small suggestion for feature branches I would prefer rebasing instead of merging to keep cleaner git log.

GediminasZukas:
OK. Thanks for advice.

PR #45, opened 2016-05-23
Change paddings and background for my profile

vycius:


RenatBuga:
1. Align icons and edit text fields center horizontally
2. Change background color to #FFF

image

PR #54, opened 2016-07-20
Show error Toast if picked location is not Vilnius

inetaa:
https://trello.com/c/fnNTOzIW/58-rework-picking-location-screens


app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
```diff
@@ -230,8 +231,16 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                     break;
                 case REQUEST_PLACE_PICKER:
                     Place place = PlacePicker.getPlace(this, data);
-                    mAddProblemLocation.setText(place.getName());
-                    locationCords = place.getLatLng();
+                    CharSequence address = place.getAddress();
+                    String matchVilnius = "Vilnius";
+                    Boolean isVilnius = Arrays.asList(address.toString().split(" ")).contains(matchVilnius);
```
vycius: This seems really hacky. Do we have any guarantees about address format ? 
Consider such example: `Gabijos 47 - 7 Vilnius, Lietuva`. Such address won't work.
I would recommend to check place coordinates if they are inside Vilnius boundary instead of depending on address string.

inetaa: Yes, it was a bit hacky. Refactored using Geocoder, seems to be more error prone although not 100% guarantee here as well. :)

app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
```diff
@@ -230,8 +236,25 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                     break;
                 case REQUEST_PLACE_PICKER:
                     Place place = PlacePicker.getPlace(this, data);
-                    mAddProblemLocation.setText(place.getName());
-                    locationCords = place.getLatLng();
+                    LatLng latLng = place.getLatLng();
+
+                    Geocoder geocoder = new Geocoder(getApplication(), Locale.getDefault());
```
vycius: Why do we need application context instead of activity context ?

vycius: If you use default locale use `new Geocoder(Context)` constructor. In this way geocoder will use default locale.

app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
```diff
@@ -230,8 +236,25 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+
+                    List<Address> addresses = null;
+                    try {
+                        addresses = geocoder.getFromLocation(latLng.latitude, latLng.longitude, 1);
+                    } catch (IOException e) {
+                        e.printStackTrace();
```
vycius, line 248: Consider logging exception to Firebase crashes.

app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
vycius: IMHO it looks good and Geocoder solution looks nearly fantastic (of course there is still small possibility that it will fail).

app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
```diff
@@ -52,44 +57,44 @@
 
 public class NewProblemActivity extends BaseActivity {
 
-    private static final int REQUEST_IMAGE_CAPTURE = 1;
+    private static final int REQUEST_IMAGE_CAPTURE   = 1;
     private static final int PERMISSION_REQUEST_CODE = 10;
 
-    public static final int REQUEST_PLACE_PICKER = 11;
-    public static final int REQUEST_PROFILE = 12;
+    public static final int REQUEST_PLACE_PICKER       = 11;
```
ViliusKraujutis: Is this some kind of auto formating configuration? Is it shared among everyone?

app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
ViliusKraujutis, line 248: What about next line?

app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
```diff
@@ -299,4 +327,17 @@ public boolean onSupportNavigateUp() {
 
         return false;
     }
+
+    private void showPlacePicker(View view){
+        PlacePicker.IntentBuilder builder = new PlacePicker.IntentBuilder();
+        try {
+            Intent intent = builder.build(this);
+            Bundle bundle = ActivityOptionsCompat.makeScaleUpAnimation(view, 0, 0, view.getWidth(), view.getHeight()).toBundle();
```
ViliusKraujutis, line 335: I like this attention to details of adding beautiful animation.

ViliusKraujutis:
LGTM :+1:

app/src/main/java/lt/vilnius/tvarkau/NewProblemActivity.java
inetaa, line 248: There weren't any Firebase logging on the first commit, added them later on. So now it's fine.

PR #60, opened 2016-08-07
About page fix

inetaa:
It got a bit messy with branches, so here is new PR with latest changes on About page


inetaa:
Reviewed & confirmed by @oleyur

PR #63, opened 2016-08-09
No connection handling

inetaa:
https://trello.com/c/GceL6YnH/81-add-toast-message-in-case-of-no-connection

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/MyProfileActivity.java
+++ b/app/src/main/java/lt/vilnius/tvarkau/MyProfileActivity.java
@@ -77,7 +77,7 @@ public class MyProfileActivity extends BaseActivity implements GooglePlusSignInI
     @Override
     public void onBackPressed() {
         if (profileFragment != null && !profileFragment.isEditedByUser()) {
-            new AlertDialog.Builder(this)
+            new AlertDialog.Builder(this, R.style.MyDialogTheme)
                     .setMessage(getString(R.string.discard_changes_title))
                     .setIcon(android.R.drawable.ic_dialog_alert)
                     .setPositiveButton(R.string.discard_changes_positive, (dialog, whichButton) ->
--- a/app/src/main/java/lt/vilnius/tvarkau/MyProfileActivity.java
+++ b/app/src/main/java/lt/vilnius/tvarkau/MyProfileActivity.java
@@ -1,11 +1,9 @@
 package lt.vilnius.tvarkau;
 
-import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v7.app.AlertDialog;
 import android.view.View;
-import android.view.inputmethod.InputMethodManager;
 import android.widget.Toast;
 
 import com.google.android.gms.auth.api.Auth;
@@ -19,6 +17,7 @@ import butterknife.OnClick;
 import lt.vilnius.tvarkau.fragments.MyProfileFragment;
 import lt.vilnius.tvarkau.utils.GoogleSignInHelper;
 import lt.vilnius.tvarkau.utils.GoogleSignInHelper.GooglePlusSignInInterface;
+import lt.vilnius.tvarkau.utils.KeyboardUtils;
 
 import static com.google.android.gms.common.SignInButton.COLOR_DARK;
 import static com.google.android.gms.common.SignInButton.SIZE_WIDE;
@@ -81,8 +80,7 @@ public class MyProfileActivity extends BaseActivity implements GooglePlusSignInI
     public void onBackPressed() {
         View view = this.getCurrentFocus();
         if (view != null) {
-            InputMethodManager imm = (InputMethodManager)getSystemService(Context.INPUT_METHOD_SERVICE);
-            imm.hideSoftInputFromWindow(view.getWindowToken(), 0);
+            KeyboardUtils.closeSoftKeyboard(this, view);
         }
         if (profileFragment != null && !profileFragment.isEditedByUser()) {
             new AlertDialog.Builder(this, R.style.MyDialogTheme)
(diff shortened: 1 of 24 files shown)


app/src/main/res/layout/no_internet.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
```
GediminasZukas, line 1: Whole layout file can be changed to something like this (take a look at first lint warning): 

```
<?xml version="1.0" encoding="utf-8"?>
<TextView xmlns:android="http://schemas.android.com/apk/res/android"
    android:id="@+id/no_internet_view"
    android:layout_width="wrap_content"
    android:layout_height="wrap_content"
    android:layout_gravity="center"
    android:drawableTop="@drawable/ic_warning"
    android:fontFamily="sans-serif-regular"
    android:paddingBottom="120dp"
    android:text="@string/no_internet"
    android:textColor="@color/black_38"
    android:textSize="18sp" />
```

I think you would need to generate new or resize `@drawable/ic_warning` vector asset to 64 dp width and height.

app/src/main/res/layout/problem_detail.xml
```diff
@@ -1,141 +1,150 @@
...
+                    android:fontFamily="sans-serif-regular"
+                    android:textColor="#DE000000"
+                    android:textSize="24sp" />
+
+                <TextView
+                    android:id="@+id/problem_status"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_alignParentRight="true"
+                    android:layout_marginTop="7dp"
```
GediminasZukas: Avoid margin `dp` values which can't be divided by 8. https://material.google.com/layout/metrics-keylines.html#

app/src/main/res/layout/problem_detail.xml
```diff
@@ -1,141 +1,150 @@
...
+                    android:text="@string/issue"
+                    android:textColor="#DE000000"
+                    android:textSize="14sp" />
+
+                <TextView
+                    android:id="@+id/problem_entry_date"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_alignParentRight="true"
+                    android:fontFamily="sans-serif-regular"
```
GediminasZukas, line 89: Lint warnings everywhere at such lines in xml files. Does it work in API level < 16? Maybe would be better to add font in  Java code?

app/src/main/res/layout/no_internet.xml
```diff
@@ -0,0 +1,25 @@
...
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/no_internet_view"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical"
+    android:paddingBottom="120dp"
+    android:gravity="center">
+
+    <ImageView
```
GediminasZukas: Content description lint warning on ImageView elements can be suppressed by this line:
`android:contentDescription="@null"`

app/src/main/res/layout/problem_detail.xml
```diff
@@ -1,141 +1,150 @@
...
+                    android:textSize="14sp" />
+
+                <TextView
+                    android:id="@+id/problem_answer"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:layout_below="@id/problem_answer_label"
+                    android:paddingTop="8dp"
+                    android:lineSpacingExtra="4sp"
+                    android:textColor="#DE000000"
```
GediminasZukas: Why not to make such hard-coded color values reusable (define once in colors.xml)?

app/src/main/java/lt/vilnius/tvarkau/MyProfileActivity.java
```diff
@@ -76,6 +79,11 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
     @Override
     public void onBackPressed() {
+        View view = this.getCurrentFocus();
+        if (view != null) {
+            InputMethodManager imm = (InputMethodManager)getSystemService(Context.INPUT_METHOD_SERVICE);
+            imm.hideSoftInputFromWindow(view.getWindowToken(), 0);
```
GediminasZukas: Possible to put these 2 lines (duplicate in NewProblemActivity) in a dedicated method and make it reusable.

app/src/main/java/lt/vilnius/tvarkau/utils/NetworkUtils.java
```diff
@@ -0,0 +1,13 @@
...
+
+import android.app.Activity;
+import android.content.Context;
+import android.net.ConnectivityManager;
+
+public class NetworkUtils {
+
+    public static boolean isNetworkConnected(Activity activity) {
+        ConnectivityManager cm = (ConnectivityManager) activity.getSystemService(Context.CONNECTIVITY_SERVICE);
+        return cm.getActiveNetworkInfo() != null;
```
GediminasZukas: I think this can be not robust solution as it checks for only available network but not real internet access. For example, this method can return `true` if we connect to wireless network but it doesn't have internet access.

app/src/main/res/layout/no_internet.xml
inetaa, line 1: Changed it and you were right, vector had to be recreated with new size.

app/src/main/res/layout/problem_detail.xml
inetaa, line 89: I would suggest to increase minSdkVersion to API 16 (Jelly beans and above). I think there would not be many Icecream users out there, don't want to concentrate on solving one use problems. What do you think?

GediminasZukas, line 89: I would support idea to increase API level to number 16 (number 15 has just 1,6% of users according to this: https://developer.android.com/about/dashboards/index.html ) but I think this question is most relevant for our PM's. @RenatBuga @povilasp

RenatBuga, line 89: Agree to increase minSdkVersion to API 16

PR #74, opened 2016-09-17
Import reports from vilnius.lt account

inetaa:


app/src/main/java/lt/vilnius/tvarkau/utils/EncryptUtils.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+
+public class EncryptUtils {
+
+    public static String encrypt(String data){
+
+        String key = "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEApdx8xnpqootgc0CyGQ83laIot1tUVTl8ORZq4z4atEbcemOWOEjXZ+ecOsHoZfmP17hmV80vusNAijuh7Utq+2kYsO51pzS6GBzh/O4ET3dxbwUA5GG43KCrKAOwurgRJkfjyihmsMGjWC/c1wzHm17agD5ZcKVL2EdYos7wGHa3gkxuxHJzG2L9tYq8OC23IpQlmpOE4Qajw/IdTil6pxe3/ur+CkJRQZdavc1aoHM8nCTXNivrZYLQV4eqhn3KHUwoTFA7BHkOhLdMnUjURQM/JovLJk83R7C5K+9EeFceNiSBuilpXyfurRCVzHyxZ1wTDiFAYfC/NKXEd42rQwIDAQAB";
```
ViliusKraujutis, line 21: :exclamation: This is unsafe! Does not look like a good idea to keep a private key in source code of a public repository :-1: 

I've asked my friend Pavel to consult for best solution. Hopefully he'll have time. Anyone has a suggestion where and how we could have this string safely stored in this repository, so others could not abuse server?

ViliusKraujutis (commented):
All looks good :+1:, except, that key :-1: . See that comment.

app/src/main/java/lt/vilnius/tvarkau/utils/EncryptUtils.java
inetaa: This is public key, not private, should be ok to have it here. :)

ViliusKraujutis:
So it's :+1: then.

inetaa:
The only issue is that it's not working. Android encryption does not match back-end openssl encryption.

app/src/main/java/lt/vilnius/tvarkau/fragments/ReportImportDialogFragment.java
```diff
@@ -105,9 +115,15 @@ public static ReportImportDialogFragment newInstance() {
             .legacyApiModule(new LegacyApiModule())
             .build()
             .inject(this);
+
+        if (getArguments().getBoolean(IS_SETTINGS_ACTIVITY)) {
```
ViliusKraujutis: Just replace with one liner: `isSettingsActivity = getArguments().getBoolean(IS_SETTINGS_ACTIVITY)`

app/src/main/java/lt/vilnius/tvarkau/entity/Problem.java
```diff
@@ -107,6 +110,8 @@ public String getId() {
         return id;
     }
 
+    public String getIdForVilniusAccount() { return idForVilniusAccount; }
```
ViliusKraujutis: Recommend using Lombok's `@Data` annotation for the class, so there would be no need to write this boilerplate code for getters and setters.

app/src/main/java/lt/vilnius/tvarkau/entity/Problem.java
```diff
@@ -26,6 +26,9 @@
     @SerializedName("docNo")
     public String id;
 
+    @SerializedName("problem_id")
+    public String idForVilniusAccount;
```
ViliusKraujutis: Can this be named better?

app/src/main/java/lt/vilnius/tvarkau/fragments/ReportImportDialogFragment.java
```diff
@@ -207,6 +208,81 @@ private boolean validateVilniusSignInputFields() {
         return emailIsValid && passwordIsValid;
     }
 
+    private void loadUserReportsFromVilniusAccount(String email, Dialog dialog) {
+
+        GetProblemsParams params = new GetProblemsParams.Builder()
+            .setStart(0)
+            .setLimit(100)
+            .setDescriptionFilter(null)
```
ViliusKraujutis: Isn't `null` a default value?

app/src/main/java/lt/vilnius/tvarkau/fragments/ReportImportDialogFragment.java
```diff
@@ -207,6 +208,81 @@ private boolean validateVilniusSignInputFields() {
...
+                            }
+                        }
+                    } else {
+                        myProblemsPreferences
+                            .edit()
+                            .putString(NewProblemActivity.PROBLEM_PREFERENCE_KEY + reportId, reportId)
+                            .apply();
+                    }
+                }
+                VilniusSignInListener listener = (VilniusSignInListener) getActivity();
```
ViliusKraujutis: Next line is pretty short, so I would not define new local variable `listener`. You can write like this, and it's still pretty good readable:
`((VilniusSignInLisetner) getActivity()).onVilniusSingIn();`

ViliusKraujutis (commented):
Looks like issue with API was updated and fixed. Other new updates also LGTM. Had some suggestions, but overall it's a :+1: and can be merged.

PS.: I think, integrating `Lombok` could be done in a separate PR.

app/src/main/java/lt/vilnius/tvarkau/entity/Problem.java
inetaa: Will change to reportId and add comment how it differs from the id above.

app/src/main/java/lt/vilnius/tvarkau/entity/Problem.java
inetaa: It's quite tempting  to use such boilerplate killer. Although googled a bit and found Jake Wharton comment that "I don't use it because it's implemented as a giant hack." ([link](https://www.reddit.com/r/androiddev/comments/2nneo7/anyone_using_project_lombok/)). Now I am confused. :)

PR #92, opened 2016-11-23
Zoom to last known user location when map is loaded

mjurkus:
Not sure if zoom level is good enough.

Reference: https://trello.com/c/SRXffUJY/143-auto-zoom-in-to-the-user-s-current-location-in-map-view


ViliusKraujutis:
lgtm :+1:

PR #94, opened 2016-11-25
Fix marker info window image loading

mjurkus:
Some refactoring and marker image loading fix.

cc @ViliusKraujutis @inetaa 

Reference: https://trello.com/c/97M2PF8v/79-thumbnails-on-map-view-are-loaded-only-on-second-click


ViliusKraujutis:
:+1:

Issue #87, opened 2016-10-22
Update Google Play services version and Place Picker


ViliusKraujutis:
Google deprecated and removed our current place picker so, it's not possible to update google-play-services without updating Place picker in NewProblemActivity.

Needs to be done at one Pull Request.

vycius:
Fixed with https://github.com/vilnius/tvarkau-vilniu/pull/96

PR #96, opened 2016-11-26
Upgrade dependencies

vycius:
Upgeaded libraries and fixed issue https://github.com/vilnius/tvarkau-vilniu/issues/87

Future considerations: 
* By introducing Kotlin https://github.com/vilnius/tvarkau-vilniu/pull/95 we would could live without Butterknife;
* Drop apt and use jack `annotationProcessor`;

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,7 +12,7 @@ buildscript {
 }
 
 plugins {
-    id "me.tatarka.retrolambda" version "3.2.5"
+    id "me.tatarka.retrolambda" version "3.4.0"
 }
 
 apply plugin: 'com.android.application'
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -6,13 +6,13 @@ buildscript {
...
         classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
-        classpath 'com.github.ben-manes:gradle-versions-plugin:0.12.0'
+        classpath 'com.github.ben-manes:gradle-versions-plugin:0.13.0'
         classpath 'io.fabric.tools:gradle:1.22.0'
     }
 }
 
 plugins {
-    id "me.tatarka.retrolambda" version "3.2.5"
+    id "me.tatarka.retrolambda" version "3.4.0"
```
vycius, line 15: Retrolambda should be replaced by jack compiler or completely dropped out on kotlin introduction

mjurkus: I'd suggest to keep it for now and once we use Kotlin replace gradually.

mjurkus (approved):

PR #97, opened 2016-11-26
Enchanted logging Timber

vycius:
Problems with old solution:
* No additional logs eg. `info` level with crash report;
* Logging in production version to console;
* Need to specify TAG for logging message;
* Timber has great formatting options with lints;


ViliusKraujutis (approved):
Great! I'm using timber on other projects. Let's use it here too. :+1:

app/src/main/java/lt/vilnius/tvarkau/utils/RemoteCrashReportingTree.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class RemoteCrashReportingTree extends Timber.Tree {
+
+    @Override
+    protected void log(int priority, String tag, String message, Throwable t) {
+        if (priority == Log.VERBOSE || priority == Log.DEBUG) {
+            return;
+        }
+
+        String messageWithTag = tag + ": " + message;
```
mjurkus, line 19: Move to `else if` branch to avoid concatenation unnecessary concatenation.

mjurkus (approved):

app/src/main/java/lt/vilnius/tvarkau/utils/RemoteCrashReportingTree.java
vycius: What do you mean ? `messageWithTag` is always needed.

mjurkus: :facepalm: yeah...

PR #99, opened 2016-11-26
Drop autodagger

mjurkus:
Dropping autodagger in favour of using `BaseActivity` and `BaseFragment` where injection is performed.

Whole dagger component is defined in `ApplicationComponent`.

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/dagger/module/NewApi.kt
+++ b/app/src/main/java/lt/vilnius/tvarkau/dagger/module/NewApi.kt
@@ -1,0 +1,12 @@
+package lt.vilnius.tvarkau.dagger.module
+
+/**
+ * @author Martynas Jurkus
+ */
+
+import javax.inject.Qualifier
+import kotlin.annotation.AnnotationRetention.RUNTIME
+
+@Qualifier
+@Retention(RUNTIME)
+annotation class NewApi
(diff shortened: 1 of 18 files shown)


app/src/main/java/lt/vilnius/tvarkau/dagger/module/NewApi.kt
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * @author Martynas Jurkus
+ */
+
+import javax.inject.Qualifier
+import kotlin.annotation.AnnotationRetention.RUNTIME
+
+@Qualifier
+@Retention(RUNTIME)
+annotation class NewApi
```
vycius, line 12: What is the purpose of this annotation ?

vycius (approved):
Nice

app/src/main/java/lt/vilnius/tvarkau/dagger/module/NewApi.kt
mjurkus: It's purpose is to distinct several non-unique objects for injection.

For example in this project Dagger providers 2 Retrofit instances, 2 OkHttpClient instances. 
One from LegacyApiModule, another from ApiModule.
Without that annotation dagger would not allow to compile.

We could create similar annotation like `OldApi`, but it's redundant in my opinion.

PR #102, opened 2016-11-27
Events tracking

vycius:
* Activities are automatically tracked by Firebase;
* Fragments are tracked  by `AnalyticsUtil::trackCurrentFragment`;
* As example for further tracking events included `AnalyticsUtil::trackViewProblem` which will give insights about problems clicked;

All these events can be seen in Firebase analytics.

For access to Firebase please write to me or @RenatBuga

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/utils/AnalyticsUtil.kt
+++ b/app/src/main/java/lt/vilnius/tvarkau/utils/AnalyticsUtil.kt
@@ -2,14 +2,18 @@ package lt.vilnius.tvarkau.utils
 
 import android.app.Activity
 import android.content.Context
+import android.os.Bundle
 import android.support.v4.app.Fragment
 import com.google.firebase.analytics.FirebaseAnalytics
+import lt.vilnius.tvarkau.entity.Problem
 
 object AnalyticsUtil {
 
     private lateinit var appContext: Context
 
     private val analytics: FirebaseAnalytics by lazy { FirebaseAnalytics.getInstance(appContext) }
+
+    private val PARAM_PROBLEM_STATUS = "problem_status"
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/lt/vilnius/tvarkau/utils/AnalyticsUtil.kt
```diff
@@ -0,0 +1,35 @@
+package lt.vilnius.tvarkau.utils
+
+import android.app.Activity
+import android.content.Context
+import android.os.Bundle
+import android.support.v4.app.Fragment
+import com.google.firebase.analytics.FirebaseAnalytics
+import lt.vilnius.tvarkau.entity.Problem
+
+object AnalyticsUtil {
```
mjurkus: I would prefer if we could make it as an implementation.

`inferface Analytics`
`class RemoteAnalytics : Analytics`

and injected using dagger.
That way we could test implementation and integration. What do you think?

That one could write in fragments tests

`verify(analytics).trackProblem(any())`

mjurkus (commented):
I'd like to discuss more about using injectable service rather than util class.

app/src/main/java/lt/vilnius/tvarkau/utils/AnalyticsUtil.kt
vycius: It depends on our testing approaches, what do we want to test, where do we want to move with tests, how are we planning to develop the app. Let's move this discussion to https://github.com/vilnius/tvarkau-vilniu/pull/103

ViliusKraujutis:
I'd say we merge this tracking improvement and refactor to injectables later. :+1:

mjurkus (approved):
Awesome!

PR #103, opened 2016-11-27
Add Robolectric fragment tests

mjurkus:
-  converting ProblemDetailFragment to KT
- adding Robolectric fragment tests

To run tests use `./gradlew testDebug`

sorry for huge commit, could not split it in smaller chunks

Going to add more test cases to ProblemDetailFragment in separate PR since this one is already a bit too big for my taste.


vycius:
I want to take a look at our testing approaches. How deep we want to test the application. Do we have a lot complex business logic ? Is testing only user flows i.e click this, click that and something happens enough ?
In our app's current situation we don't have any tests. We are still iterating quite quickly with the app and it changes frequently. Of course tests help to reduce bugs number, but we really need to think that it slows development process. We need to maintain tests and app's code.
Before merging this let's answer few questions:
1) Why did you chose Roboeletric over Espresso or over test frameworks ?
2) Let's take a look at our current crashes on `Firebase` or `Crashlytics`. How similar tests will help us to increase crash-free users number ?
3) Is TDD good for everyone ?
4) On what testing approaches are we planning to rely on ?
5) What do we want test?

mjurkus:
Good questions @vycius 
`We are still iterating quite quickly with the app and it changes frequently`
Just because of this statement I'd say tests are mandatory. How do you know, that you didn't break anything when adding or removing a line? Sure you can test manually, but that takes a lot of time to check all possible cases. And in some cases not possible at all without manually changing code and restarting the app to check error handling.
Just from the top of my head - ProblemDetailsFragment - with network, without, with correct response, with malformed response. In many cases you can't even test these conditions without some steps in setting up your device.
And in some cases unit tests provide logic behind code. At the moment, in ProblemDetailsFragment if there is ANY exception we show notification that there was a network problem. Which is not true. Why is that?
Sure, this app is small. But who is to say, that it always will be that small? At my current company we seriously started adding tests just very recently. And project is now huge. And sometimes, when I have to make a change in some old part I'm afraid because nobody remembers how exactly it should work. So  when you change something and you see tests break the feeling of security is wonderful :)

1. Robolectric tests are much much faster and easier to maintain from my experience. Although both of them can be combined for maximum effect.
2. Maybe we won't have drastic decrease in crash-free sessions now, but tests can help us protect from possible bugs. What is crash free session percentage in Fabric?
3. TDD is hard. I don't suggest code using TDD approach. Just add tests after code is implemented and that is good enough.
4. What do you mean?
5. Caching (future), network (un)availability and how app behaves then, view visibility, models.

Maybe I'm a bit old fashioned that way, but I don't trust code that has no tests :)

mjurkus:
Even more, I'd like to introduce project to MVP pattern to separate responsibilities. But that could be done later.

mjurkus:
@ViliusKraujutis @inetaa any input from you?

inetaa:
Hi! I am in Sweden (work trip) for a week, so did not have much time to take a look into this in detail. Anyway, the quick comment would be that I would love to try MVP pattern and regarding tests I agree that tests provide security and stability but this project is still small and still evolving, so I would suggest to cover with tests just the most sensitive areas or areas which have issues at the moment. Otherwise it could slow the development process and that could be especially not good for a volunterring-based project as people who contribute mostly have quite limited time.
On the other note, maybe anyone have tried [ Firebase Espresso test recorder](https://developer.android.com/studio/test/espresso-test-recorder.html)? Still in beta but promises to be an easy way to write tests.

vycius:
@inetaa Yes, I tried Espresso test recorder. Well it get's the job done by creating scenario tests, but generated tests are very ugly and when you are changing places that affects test case you need to completely regenerate the test. Test generation works by clicking app on test recorded and test code it automatically generated. So it's really quick process. 
If we want to test only perfect scenarios Espresso test recorder is good enough. If we want to test not perfect user-flows e.g request fails, different responses and etc. we need to look for other possibilities.

mjurkus:
So to sum up is to use MVP, but no tests?

inetaa:
I think that would be the best decision for the current project state. And we can always come back to this discussion later and add tests in the future. :)

ViliusKraujutis:
Did not look into details much, but tests are always a good idea. 
I don't see it would be a breaker for future development - only benefits, so @mjurkus, I give :+1: for this PR.

mjurkus:
OK. @vycius @inetaa here is my suggestion.
We won't ask contributors to add tests to accept PRs. Lets leave it optional. Those who want to add tests lets allow them.
How about that?

vycius:
@mjurkus I don't think that anyone is saying that tests is not beneficial thing. 
The main part of the question is to decide which testing approaches we choose that everyone should be happy about it and write at least simple tests with infrastructure created by this PR. 
We don't want to end up with Roboeletric, Espresso and few other testing frameworks in one project. 

I would be in favor of having tests with Roboelectric as @mjurkus proposed. One of the requirements that tests should pass to merge with mater. With new functionality it's not required, but really recommended  to write tests for that. 
Also, I would like to have some tool to test application logic which is not covered by tests. E.g introduce Robo tests. It would randomly click on our application UI elements and see if there is no application crashes. We can run something like this on every PR.

app/build.gradle
```diff
@@ -95,13 +95,13 @@ ext {
     parcelerVersion = '1.1.6'
     threeTenVersion = '1.0.4'
     photoView = '1.3.1'
+    robolectric_version = '3.1.2'
```
vycius: Use same style e.g `robolectricVersion`, `kotlinMockitoVersion`

mjurkus:
So. Like we already established we need 1 Approved or 1 LGTM :)

app/build.gradle
```diff
@@ -120,7 +120,7 @@ dependencies {
     provided "frankiesardo:icepick-processor:$icepickVersion"
 
     compile "com.jakewharton:butterknife:$butterKnifeVersion"
-    annotationProcessor "com.jakewharton:butterknife-compiler:$butterKnifeVersion"
+    kapt "com.jakewharton:butterknife-compiler:$butterKnifeVersion"
```
vycius, line 123: What is the difference beetween `annotationProcessor` and `kapt` ? Is `kapt` already stable ?

app/src/test/java/lt/vilnius/tvarkau/base/BaseRobolectricTest.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+@RunWith(RobolectricTestRunner::class)
+abstract class BaseRobolectricTest {
+
+    protected lateinit var activity: TestActivity
+    protected var activityController: ActivityController<TestActivity>? = null
+
+    @Before
+    open fun setUp() {
+        ShadowApplication.getInstance().declareActionUnbindable("com.google.android.gms.analytics.service.START")
```
vycius: Why ?

app/build.gradle
```diff
@@ -158,10 +158,27 @@ dependencies {
         transitive = true;
     }
 
-    testCompile "org.robolectric:robolectric:3.1.4"
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-
     compile 'com.jakewharton.timber:timber:4.3.1'
+
+    kaptTest "com.google.dagger:dagger-compiler:$daggerVersion"
```
vycius, line 164: No need to define dependency second time. It should be used from regular `kapt`

app/build.gradle
```diff
@@ -158,10 +158,27 @@ dependencies {
...
+    testCompile "org.jetbrains.kotlin:kotlin-test:$kotlin_version"
+    testCompile "junit:junit:$junitVersion"
+    testCompile "org.robolectric:robolectric:$robolectric_version"
+    testCompile "org.robolectric:shadows-support-v4:$robolectric_version"
+    testCompile "org.jetbrains.kotlin:kotlin-test:$kotlin_version"
+    testCompile "com.google.dagger:dagger:$daggerVersion"
```
vycius, line 170: Same. Duplicated dependency

app/src/main/java/lt/vilnius/tvarkau/BaseActivity.java
```diff
@@ -17,7 +17,7 @@
     @Inject
     SharedPreferences myProblemsPreferences;
 
-    private ApplicationComponent component;
+    protected ApplicationComponent component;
```
vycius, line 22: If access level was changed only for testing purposes add `VisibleForTesting` annotation. For instance: 
```java
@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```

app/src/test/java/lt/vilnius/tvarkau/fragments/ProblemDetailFragmentTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    override fun setUp() {
+        super.setUp()
+        activity.getTestComponent().inject(this)
+    }
+
+    @Test
+    fun problemWithoutAnswer_displayCorrectLayout() {
+        val problem = Problem()
+        whenever(api.getProblem(any())).thenReturn(problem.wrapInResponse)
```
vycius, line 41: What about simulating network failures e.g receiving status 500 from the server or simulating error codes returned in response ? What do you think about creating mock Retrofit client and using actual responses from the server stored in resource files ?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.1-all.zip
```
vycius: Is Gradle 3.1 officially supported with Android studio 2.2.2. Why not the newest one 3.2.1 ?

vycius: Also it's major release so I would recommend to regenerate the wrapper

vycius (changes requested):
You forgot to run tests on Travis 😿 
Left some comments, but overall everything looks good except that we need to run these tests somewhere.

app/build.gradle
mjurkus: kapt is kotlin annotation processor. It knows how to read annotations in kotlin and java classes. APT on the other hand doesn't.

mjurkus: And yes, it's stable. Kapt2 is already in beta.

app/build.gradle
mjurkus: `kapt` vs `katpTest` is key difference.
`kapt` only runs on `src/main`
`katpTest` only on `src/test`

app/src/test/java/lt/vilnius/tvarkau/base/BaseRobolectricTest.kt
mjurkus: Ups. Copy paste from other project.
Basicly it disables services if there are any.

gradle/wrapper/gradle-wrapper.properties
mjurkus: How?

mjurkus: Maybe not officially, but it works without any issues.

vycius: With `gradle wrapper`

app/src/test/java/lt/vilnius/tvarkau/fragments/ProblemDetailFragmentTest.kt
mjurkus: I've tried using resources for actual responses in previous projects, but it's areal hassle to keep them up to date.
It's much easier just to create entities manually as needed.
Otherwise you must have heaps of files with different amounts of data.
For example, just with problem, with problem and answer, different problem types, failure responses, validation responses and so on.

vycius:
Also if possible update Readme with some information about testing.

app/src/main/java/lt/vilnius/tvarkau/BaseActivity.java
mjurkus: Somehow it doesn't work for me. Am I doing it wrong?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Aug 17 09:08:57 EEST 2016
+#Tue Nov 29 20:25:18 EET 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.2.1-bin.zip
```
vycius, line 6: I see some warnings from Gradle on Travis build https://travis-ci.org/vilnius/tvarkau-vilniu/builds/179856849 but these due to deprecation so it should be ok

vycius (approved):
👍 LGTM

app/src/test/resources/robolectric.properties
```diff
@@ -0,0 +1,3 @@
+packageName=lt.vilnius.tvarkau
+sdk=23
```
vycius, line 2: Why 23 ? Can't we use newest version 25 ?

vycius: Is it possible to also run tests on our oldest supported sdk version ?

mjurkus: Not supported by Robolectric: https://github.com/robolectric/robolectric/issues/2554

PR #104, opened 2016-11-29
Migrate Problem entity to KT

mjurkus:
Migrate entity, add some tests for more complicated parts.


app/src/main/java/lt/vilnius/tvarkau/entity/Problem.kt
```diff
@@ -0,0 +1,52 @@
+package lt.vilnius.tvarkau.entity
+
+import com.google.android.gms.maps.model.LatLng
+import com.google.gson.annotations.SerializedName
+import org.parceler.Parcel
+import org.threeten.bp.LocalDateTime
+
+@Parcel
+data class Problem(
```
vycius, line 9: I would prefer to have separate classes for `Problem` and `ProblemResponseDto` with `ProblemMapper`. In this way inside `Problem` class we wouldn't have unneeded fields for example `typeName` and `type` would become `type` enum, `entryDate` and `reportDate` would become `entryDate`, `photo` and `thumbnail` would be `photos` ...

app/src/main/java/lt/vilnius/tvarkau/entity/Problem.kt
```diff
@@ -0,0 +1,52 @@
...
+            return photo ?: thumbnail?.let { listOf(it) }
+        }
+
+    companion object {
+        const val STATUS_DONE = "Atlikta"
+        const val STATUS_RESOLVED = "Išnagrinėta"
+        const val STATUS_TRANSFERRED = "Perduota"
+        const val STATUS_REGISTERED = "Registruota"
+        const val STATUS_POSTPONED = "Atidėta"
+    }
```
vycius, line 51: I would prefer to have type enum. In this way `Problem.applyReportStatusLabel` would look nice, problem type would be completely type safe, less error prone.

vycius (approved):
👍 Looks far better. Left few comments to think about.

ViliusKraujutis (approved):

app/src/main/java/lt/vilnius/tvarkau/entity/Problem.kt
mjurkus: backend should provide better data to fully utilise enums.
Now we would still have to map `Atidėta` to enums, so no much win.

ViliusKraujutis: Enums can have 
```java
@SerializeName("Atidėta")
```

app/src/main/java/lt/vilnius/tvarkau/entity/Problem.kt
mjurkus: DTO pattern is good, but often missused by creating layers and layers of duplicate data objects.
IMO DTO are good when it's used with databases and filled selectively with only required data.
I don't think this would be a good use for it. Yes, we have extra fields, but now (with this) implementation it is hidden from end user.

app/src/main/java/lt/vilnius/tvarkau/entity/Problem.kt
mjurkus: Yeah, but what are cons to have additional mapping?
```
enum State {
  @SerializeName("Atidėta")
  DELAYED,
  ...
}
```

With such mapping you'd still need maping back to LT translation to display to user.

PR #109, opened 2016-12-01
Fix NPE when opening ProblemsMapActivity

mjurkus:
Fix issue and add tests

Reference: https://trello.com/c/qQhZzX4W/150-bug


vycius (approved):
LGTM

PR #110, opened 2016-12-01
Fix opening of full screen photo activity

mjurkus:
Ref: https://trello.com/c/No4Xi8GO/151-bug-photo-full-screen-view-doesn-t-work


vycius (approved):
Looks good

PR #114, opened 2016-12-02
Update versionCode to 17, versionName to 3.4

mjurkus:


vycius (approved):
I hope we won't need to do similar PR's in the future by automating release process.

PR #115, opened 2016-12-03
Auto publishing to Google play

vycius:
New parameters for Gradle build `versionCode` and `versionName`:
```gradle
 ./gradlew -Pversion_code=4 -Pversion_name=1.3 assembleDebug
```
Push tag with `versionName` and it will be automatically uploaded to Google play Beta track. `versionCode` is used from `Bitrise build number`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,5 +26,23 @@ To run tests from command line run `./gradlew testDebug`
 
 To run tests from Android Studio select test class or package, right click and select `Run Tests`
 
+## Versioning
+Fluid versioning is used for this project.
+
+The number in the first position (**1**.2.3.4) indicates a **major release**.
+
+The number in the second position (1.**2**.3.4) indicates a **minor release**.
+
+The number in the third position (1.2.**3**.4) indicates a **bug fix release**.
+
+The number in the fourth position (1.2.3.**4**) indicates a **special micro release**.
+
+Version numbers should be changed in `app/build.gradle` file.
+
+## Publishing
+
+After merging pull request with master branch app will be automatically published to Google play Beta track and tag with version name will appear in GitHub.
+
+
 ### Note for Mac/Linux users
 follow small guide at http://robolectric.org/getting-started/ section `Note for Linux and Mac Users`
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,6 +17,11 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 apply plugin: 'com.github.ben-manes.versions'
 
+def versionMajor = 3
+def versionMinor = 5
+def versionPatch = 0
+def versionMicro = 0
+
 android {
     compileSdkVersion 25
     buildToolsVersion "25.0.1"
@@ -25,8 +30,8 @@ android {
         applicationId "lt.vilnius.tvarkau"
         minSdkVersion 16
         targetSdkVersion 25
-        versionCode 17
-        versionName "3.4"
+        versionCode versionMajor * 1000000 + versionMinor * 10000 * versionPatch * 100 + versionMicro
+        versionName "$versionMajor.$versionMinor.$versionPatch.$versionMicro"
         vectorDrawables.useSupportLibrary = true
     }
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -176,6 +176,12 @@ dependencies {
     testCompile 'org.khronos:opengl-api:gl1.1-android-2.1_r1'
 }
 
+task versionName() << {
+    description = "Prints version name"
+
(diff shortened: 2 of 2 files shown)


mjurkus:
I'm not sure this is a viable solution. We shouldn't do releases from master branch...
Could we discuss this at next meetup?

vycius:
Yes of course, but I won't be in the next meetup. Although this solution works perfectly with release branches. You push release branch with a tag and `.apk` ends up Google play Beta track. So what do you mean not viable ?

mjurkus:
A lot of room for human error. One could easily make mistake creating a build.
From codebase can't see which version has what features.
How it will work with Bitrise?

vycius:
All released versions automatically appear under GitHub releases. By pushing tag you should supply description e.g what has been fixes and It contains link to commit ... For instance it would look like this https://github.com/JakeWharton/timber/releases In this way you can easily keep track on which versions have been released, what has been fixed, which features each version have.
From Bitrise perspective it will have git trigger which triggers for each pushed tag event. It will automatically run tests, build apks, sign them and if everything succeeds upload signed `.apk` to Google play Beta track. On successful upload message containing description, version will be sent to Slack. 
@mjurkus What is your opinion about best way to publish to Google play ?

vycius:
@ViliusKraujutis @inetaa What is opinion about this PR ?

inetaa:
I agree with @mjurkus that auto publishing by tag may lead to some human errors (not that they would be very crucial). But it's hard to know which versionName to push and then versionCode is unknown from Bitrise which as well may lead to some confusion if we use Analytics or when we add same versionCode to app for users to understand which version they have. I would suggest to make publishing available to anyone (not just developers), so for this approach every new commit after merge should be release ready (with updated version name & code) and from Bitrise anyone could just push the button. I don't have logins to Bitrise but I guess they are quite similar as Buddybuild which I currently use. Adding tag to release commit is nice for the history reference, we could do that as well.

vycius:
@inetaa Are suggesting to publish to Beta track on every Pull requests acceptance ?

ViliusKraujutis:
I think, it's a great idea to push nightly builds to Google Play Alpha release. I'd love to have some script, which would bump version code in let's say `versions.gradle` file. Then from Google Play Developer Console, from Alpha Tab, even @RenatBuga would be able to promote a certain version to Beta channel or even straight to Production.
Google Play Developer Console Alpha tab screenshot

PS.: Also, I've checked, that we have just 500+ commits per year, so I even don't see a problem, if we'd push Alpha build after every merge to master branch.

mjurkus:
I do agree about publishing to alpha/beta, but I can't agree with fluid versioning.
I suggest to use plugin, properties file and gradle commands to increment app version.
I do have a plugin for that, but it's currently in private repo... Maybe there are some smarter ones somewhere on GH?

ViliusKraujutis:
What do you mean by `fluid versioning`? You mean you don't agree to have 4 positions for versioning? :+1: Same here.

Here's a doc where I've familiarized with term `fluid versioning`, so we'd be talking about same thing https://wiki.fluidproject.org/display/Infusion13/Fluid+Versioning+Scheme

mjurkus:
No, I mean that app version is not visible in the project anywhere.
Actually I do agree usging 4 number for versioning. `1.2.3.4`.
This defines clear scheme for app versioning.
1 - major, almost never changes
2 - minor, changes with new feature release to production channel
3 - patch, when release of quick fixes, like crashes to production
4 - beta, for internal builds

vycius:
My vision about auto publishing to beta / alpha on every merge. 
IMHO version name is only important for end users (when version reaches prod). For internal builds version code is far important. 
@mjurkus @ViliusKraujutis  I don't think that we need version name consisting of 4 numbers. Could you provide details what would be use case for fourth number ? Currently I think 3 numbers is more than enough e.g `a.b.c`:
* a - major version;
* b - minor version;
* c - patch version;

These can be defined in `build.gradle` file as:
```groovy
def versionMajor = 1
def versionMinor = 2
def versionPatch = 3
```
Even @RenatBuga would be able to change this straight from the Github and trigger publish to Beta / alpha channels. 

Version code building:
Version code shouldn't be just number that is increased by one on every build. This number is really important for us, because we see it in crashlogs, analytics ... My suggestion is to build version code from version number and include build number to increase it on every publish. At the end it might look something like this:
`abbccnnn`, here `a` - major version, `b` - minor version, `c` - patch version, `n` - build number in our CI. 

@mjurkus @inetaa @ViliusKraujutis What would you think about such approach ?

vycius:
At second thought I think that fluid versioning would work correctly. Just we need to build `version_code` from it.

mjurkus (approved):

PR #116, opened 2016-12-04
Log fragment open as separate event

vycius:
Unfortunately `setCurrentScreen` is not logged as separate event and it doesn't show in the funnel. Make fragment open as separate event to show it in the funnel.

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/analytics/RemoteAnalytics.kt
+++ b/app/src/main/java/lt/vilnius/tvarkau/analytics/RemoteAnalytics.kt
@@ -14,9 +14,18 @@ class RemoteAnalytics(appContext: Context) : Analytics {
     }
 
     private val PARAM_PROBLEM_STATUS = "problem_status"
+    private val PARAM_FRAGMENT_ACTIVITY = "activity"
 
     override fun trackCurrentFragment(activity: Activity, fragment: Fragment) {
-        analytics.setCurrentScreen(activity, fragment.javaClass.simpleName, null)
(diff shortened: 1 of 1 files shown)


app/src/main/java/lt/vilnius/tvarkau/analytics/RemoteAnalytics.kt
```diff
@@ -14,9 +14,18 @@ class RemoteAnalytics(appContext: Context) : Analytics {
     }
 
     private val PARAM_PROBLEM_STATUS = "problem_status"
+    private val PARAM_FRAGMENT_ACTIVITY = "activity"
 
     override fun trackCurrentFragment(activity: Activity, fragment: Fragment) {
-        analytics.setCurrentScreen(activity, fragment.javaClass.simpleName, null)
+        val fragmentName = fragment.javaClass.simpleName
+
+        Bundle().run {
```
mjurkus, line 22: This `run {}` really doesn't help to readability or give any other benefit.
```
val bundle = Bundle()
bundle.putString(PARAM_FRAGMENT_ACTIVITY, activity.javaClass.simpleName)
analytics.logEvent("open_$fragmentName", bundle)
```
looks equally good.

mjurkus (approved):

app/src/main/java/lt/vilnius/tvarkau/analytics/RemoteAnalytics.kt
vycius: Personally I prefer version with `run` in this case.

PR #117, opened 2016-12-04
Map in problem details

vycius:
screenshot_20161206-212355


app/src/main/java/lt/vilnius/tvarkau/fragments/ProblemDetailFragment.kt
```diff
@@ -127,13 +129,9 @@ class ProblemDetailFragment : BaseFragment() {
...
 
-                        problem.photos?.let {
-                            if (it.size == 1) {
-                                problem_images_view_pager_indicator.gone()
-                            }
-
-                            initProblemImagesPager(problem)
-                        } ?: problem_image_pager_layout.gone()
+                        if (problem.photos == null || problem.photos!!.count() == 1) {
+                            problem_images_view_pager_indicator.gone()
```
mjurkus: Can use `goneIf()` extension

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+        if (isEmpty()) {
+            return 1
+        }
+
+        return photos.count() + if (problem != null) 1 else 0
+    }
+
+    fun isEmpty(): Boolean = photos.isEmpty() && problem == null
+
+    override fun isViewFromObject(view: View, `object`: Any): Boolean {
```
mjurkus: `object` > `any`

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+        return view === `object`
```
mjurkus: inline

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        container.addView(itemView)
+
+        return itemView
+    }
+
+    fun instantiateProblemMap(container: ViewGroup): View {
+        layoutInflater.inflate(R.layout.problem_images_view_pager_map, container, false).run {
+            if (problem_map_view != null) {
```
mjurkus: when `problem_map_view` could be null?

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+                problem_map_view.onCreate(null)
+                problem_map_view.getMapAsync {
+                    it.run {
```
mjurkus: What is this additional `run` is for?

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+                        moveCamera(CameraUpdateFactory.newLatLng(problem!!.latLng))
```
mjurkus: If someone were to remove if condition in `instantiateItem` this would throw exception. Suggest passing problem directly to this function.

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+                }
+            }
+            return this
+        }
+    }
+
+    fun instantiateEmptyPhoto(container: ViewGroup): View =
+            layoutInflater.inflate(R.layout.no_image, container, false)
+
+
```
mjurkus, line 61: empty line

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+    fun instantiateProblemPhoto(position: Int, container: ViewGroup): View {
+        layoutInflater.inflate(R.layout.problem_images_view_pager_item, container, false).run {
+            val photo = photos[position]
+
+            Glide.with(context).load(photo).into(problem_image_view)
+
+            problem_image_view.setOnClickListener { v ->
```
mjurkus: `v` is not used

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,109 @@
...
+import lt.vilnius.tvarkau.R
+import lt.vilnius.tvarkau.entity.Problem
+
+
+class ProblemImagesPagerAdapter(context: Context, private val photos: List<String>) : PagerAdapter() {
+
+    private val layoutInflater: LayoutInflater
+    private var problem: Problem?
+
+    constructor(context: Context, problem: Problem) : this(context, problem.photos.orEmpty()) {
```
mjurkus: I'd suggest using 2 adapters, since it now breaks SRP.

mjurkus (changes requested):

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
vycius: Now using two adapters

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+
+        container.addView(itemView)
+
+        return itemView
+    }
+
+    fun instantiateProblemMap(mapView: View): View {
+        mapView.problem_map_view.onCreate(null)
+        mapView.problem_map_view.getMapAsync {
+            it.run {
```
mjurkus: Why do you use `run { }` here?

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+    override fun getCount(): Int = problem.photos.orEmpty().count() + 1
+
+    override fun isViewFromObject(view: View, any: Any): Boolean = view === any
+
+    override fun instantiateItem(container: ViewGroup, position: Int): Any {
+        val layoutInflater = container.context.getSystemService(Context.LAYOUT_INFLATER_SERVICE) as LayoutInflater
+
+        val itemView: View
+
+        if (position == count - 1) {
```
mjurkus, line 30: Why + 1 at line 21 and - 1 here?

mjurkus (commented):
Still I would suggest refactoring `ProblemImagesPagerAdapter` even further.
`ProblemImagesPagerAdapter` should not contain logic on what it should display.
Logic on wether it should show map or photo should be contained In fragment of activity where this adapter is used.
Because now adapter is still violating SRP. It's name even states that it displays `ProblemImages`

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
vycius: +1 is for displaying map at the last image splot

app/src/main/java/lt/vilnius/tvarkau/views/adapters/ProblemImagesPagerAdapter.kt
vycius: Why do you think `run { }` is misused here ?

mjurkus: Because `run { }` does not provide any value.

mjurkus:
Nevermind that long comment about splitting adapter even further. From adapter's name I thought it is used to display photos OR map.

vycius:
Yup I think name `ProblemImagesPagerAdapter` is confusing, but I haven't come up with better name. It displays photos and lite map (like photo of the map). Do you have any suggestions for naming here ?

inetaa:
Regarding naming it was decided some time ago not to use _Problem_ at all because of negative aspect, instead please use _Report_.

vycius:
@inetaa we should do renaming in whole project. It will be really confusing if only in this adapter we will use report and in another places problem.

mjurkus (approved):

PR #118, opened 2016-12-06
Cleanup libraries

vycius:
Removed few unused libraries and replaced `com.android.support:support-v4` by  `com.android.support:support-fragment`. 
With this PR .apk size drops about 0.2 MB and method count decreases about 2K.
More statistics about libraries we use
screen shot 2016-12-06 at 22 06 49

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -28,6 +28,7 @@
     <color name="snackbar_action_text">#3bdfff</color>
 
     <color name="black_38">#9E9E9E</color>
+    <color name="black_87">#de000000</color>
     <color name="problem_date">#6d6d6d</color>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/colors.xml
```diff
@@ -28,6 +28,7 @@
     <color name="snackbar_action_text">#3bdfff</color>
 
     <color name="black_38">#9E9E9E</color>
+    <color name="black_87">#de000000</color>
```
mjurkus, line 31: Why new color?

vycius: This color is used from `com.github.ParkSangGwon:TedPicker` library. So extracted it `colors.xml` and removed this dependecy

ViliusKraujutis (approved):

PR #120, opened 2016-12-06
Use default android copy mechanics

mjurkus:
After discussion @ C4V meetup @inetaa showed how wrong we are in our ways and proved, that default copy is better than using custom solution 🙇 😄 

Ref: https://trello.com/c/X9IALIkJ/163-replace-the-auto-copy-function-with-regular-selector-on-long-tap


vycius (approved):
I am interested in how she proved this :D ?

mjurkus:
Showed that people swipe the text to select it rather than long clicking it to invoke default selection function.

PR #121, opened 2016-12-07
Initial flow of new report registration

mjurkus:
No MVP yet. Too much boilerplate for such small feature IMO.


app/src/main/res/layout/item_report_type.xml
```diff
@@ -1,26 +1,28 @@
...
+              android:layout_width="match_parent"
+              android:layout_height="wrap_content"
+              android:orientation="vertical">
 
     <TextView
         android:id="@+id/item_report_type_name"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
+        android:background="?attr/selectableItemBackground"
+        android:clickable="true"
```
mjurkus, line 12: Allows ripple effect

app/src/main/java/lt/vilnius/tvarkau/fragments/ReportTypeListFragment.kt
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+
+        val request = ApiRequest<GetProblemTypesParams>(ApiMethod.GET_PROBLEM_TYPES, null)
+
+        legacyApiService.getProblemTypes(request)
+                .toSingle()
+                .doOnSubscribe { report_types_progress.visible() }
```
ViliusKraujutis, line 63: Can we have ID naming convention changed from snake to camel case? So this would look like `reportTypesProgress` instead of `report_types_progress`?

ViliusKraujutis (approved):

app/src/main/java/lt/vilnius/tvarkau/fragments/ReportTypeListFragment.kt
mjurkus: No. This way it's clear just from seeing the code, that this is a view and not a class property.

vycius (approved):

PR #124, opened 2016-12-09
Migrate About screen to Kotlin and show only version name

vycius:


app/src/main/java/lt/vilnius/tvarkau/AboutActivity.kt
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            startActivity(intent)
+        } catch (e: ActivityNotFoundException) {
+            startActivity(Intent(Intent.ACTION_VIEW,
+                    Uri.parse("http://play.google.com/store/apps/details?id=$packageName")))
+        }
+
+    }
+
+    @OnClick(R.id.contribute_with_code)
```
mjurkus: You want to use KT and Butterknife?

vycius: Used KT for eveythink

PR #125, opened 2016-12-10
Rewrite taking photos

vycius:
* Use EasyImage library to take photos and select photos from gallery. We use this library in Trafi;
* Old `createBitmap` logic was very complex and hacky. Rewritten it using Glide.


mjurkus (approved):
Nice

mjurkus:
118 pts of karma for you

PR #128, opened 2016-12-11
Unsubscribe from observables when views are destroyed

vycius:
There is plenty of exceptions similar to this https://console.firebase.google.com/project/tvarkau-vilniu/monitoring/app/android:lt.vilnius.tvarkau/cluster/7c183202?duration=2592000000&appVersions=3050200

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/fragments/MultipleProblemsMapFragment.kt
+++ b/app/src/main/java/lt/vilnius/tvarkau/fragments/MultipleProblemsMapFragment.kt
@@ -12,6 +12,7 @@ import lt.vilnius.tvarkau.backend.ApiMethod
 import lt.vilnius.tvarkau.backend.ApiRequest
 import lt.vilnius.tvarkau.backend.GetProblemsParams
 import lt.vilnius.tvarkau.entity.Problem
+import rx.Subscription
 import rx.android.schedulers.AndroidSchedulers
 import rx.functions.Action1
 import rx.schedulers.Schedulers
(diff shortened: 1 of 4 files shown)


app/src/main/java/lt/vilnius/tvarkau/fragments/MultipleProblemsMapFragment.kt
```diff
@@ -54,7 +58,7 @@ class MultipleProblemsMapFragment : BaseMapFragment(),
             Timber.e(throwable)
         }
 
-        legacyApiService.getProblems(request)
+        subscription = legacyApiService.getProblems(request)
```
mjurkus, line 61: nicer way IMO is
```
  .....
  .subscribe()
  .apply{ subscription = this }
```

mjurkus (approved):

app/src/main/java/lt/vilnius/tvarkau/fragments/MultipleProblemsMapFragment.kt
vycius: I would prefer current style as you see right away if it's assigned to subscription or not.

mjurkus: But your way breaks nice flow of Rx.

For example if there were multiple possible subscriptions one could use
```
  .....
  .subscribe()
  .apply{ subscriptions += this }
```

PR #138, opened 2016-12-21
Fill date time field with exif date time if possible

mjurkus:
Reference: https://trello.com/c/4CWUGCU8/158-add-options-to-insert-info-into-date-time-field

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/utils/FormatUtils.java
+++ b/app/src/main/java/lt/vilnius/tvarkau/utils/FormatUtils.java
@@ -1,5 +1,6 @@
 package lt.vilnius.tvarkau.utils;
 
+import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.text.Spannable;
 import android.text.SpannableString;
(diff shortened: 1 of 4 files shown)


app/src/main/java/lt/vilnius/tvarkau/utils/FormatUtils.java
```diff
@@ -13,6 +14,12 @@
 
 public class FormatUtils {
 
+    private static final DateTimeFormatter EXIF_DATE_TIME = DateTimeFormatter.ofPattern("yyyy:MM:dd HH:mm:ss");
+
+    public static String formatExifAsLocalDateTime(@NonNull String exifDateTime) {
```
vycius, line 24: Make this method to return null on exception. From function definition we don't know if it can throw an exception.

vycius (changes requested):

app/src/main/java/lt/vilnius/tvarkau/utils/FormatUtils.java
mjurkus: Then maybe always null?

vycius: Why always ? I see no reason to have try/catch in `NewReportFragment.kt` just because this function might throw exception.

mjurkus: I mean try/catch in `formatExifAsLocalDateTime` and on error return null

vycius: Yup. Just move `NewReportFragment.kt` try/catch to `formatExifAsLocalDateTime`.

mjurkus:
Updated

vycius:
I think one problem exists with this PR. If user selected date from calendar manually after selecting photo, date is overridden by photo EXIF date. In this case user should reselect date from calendar.

mjurkus:
@vycius we already discussed this with @RenatBuga. It should override any date already entered in that field.

vycius (approved):
LGTM

PR #139, opened 2016-12-21
When requesting permissions from fragment don't use ActivityCompat

mjurkus:
Result was returned to activity instead of fragment.

Reference: https://trello.com/c/EAYd63Qw/169-bug-user-must-repeat-action-after-permissions-granted


vycius (approved):
LGTM. Nice catch.

PR #140, opened 2016-12-22
Implement preferences provider

mjurkus:
### Usage

Inject using dagger like so:
```java
@field[Inject Named(Preferences.DISPLAY_PHOTO_INSTRUCTIONS)]
lateinit var displayPhotoHints: BooleanPreference
```

Check or update values
```
displayPhotoHints.get()
displayPhotoHints.set(newValue)
```

Easy to use, easy to mock.


app/src/main/java/lt/vilnius/tvarkau/dagger/module/SharedPreferencesModule.kt
```diff
@@ -0,0 +1,32 @@
...
+        return application.getSharedPreferences(MY_PROBLEMS_PREFERENCES, Context.MODE_PRIVATE)
+    }
+
+    @Provides
+    @Singleton
+    @Named(DISPLAY_PHOTO_INSTRUCTIONS)
+    fun providePhotoInstructions(
+            preference: SharedPreferences
+    ): BooleanPreference {
+        return BooleanPreferenceImpl(preference, DISPLAY_PHOTO_INSTRUCTIONS, true)
```
vycius, line 30: What is the purpose of `PreferenceImpl` abstraction ? Why it's not good enough to access shared preferences directly from `SharedPreferences` ?

mjurkus: I just find it much more convenient when using multiple fixed name preferences.
Instead of `sharedPreferences.edit(). putBoolean("key", false).commit()` in multiple places one can use `displayPhotoHints.set(false)`.
Furthermore `BooleanPreference` is mockable and it's much easier to setup tests.

app/src/main/java/lt/vilnius/tvarkau/prefs/IntPreferenceImpl.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import android.content.SharedPreferences
+import lt.vilnius.tvarkau.prefs.BasePreferenceImpl
+import lt.vilnius.tvarkau.prefs.IntPreference
+import lt.vilnius.tvarkau.prefs.Serializer
+
+/**
+ * @author Martynas Jurkus
+ *
+ * @since 2014-03-24
```
vycius: Remove misleading since

vycius (approved):
After some thinking this solution for preferences looks awesome. 
I would like to see such solution implemented for all our preferences.

PR #143, opened 2016-12-27
Implement tracking events

mjurkus:
Reference: https://trello.com/c/0U189FG5/167-set-tracking-of-events


app/src/main/java/lt/vilnius/tvarkau/analytics/RemoteAnalytics.kt
```diff
@@ -37,4 +35,45 @@ class RemoteAnalytics(appContext: Context) : Analytics {
         analytics.logEvent(FirebaseAnalytics.Event.VIEW_ITEM, this)
     }
 
+    override fun trackReportRegistration(reportType: String, photoCount: Int) {
+        val params = Bundle().apply {
+            putString(PARAM_REPORT_TYPE, reportType)
```
vycius: Firebase `CONTENT_TYPE` can be used here.

vycius (approved):
LGTM 
In the future we should introduce user property for is user logged in.

PR #146, opened 2016-12-29
Separate my problems preferences and other preferences

mjurkus:
Forgot to separate preferences in initial implementation.

Issue was that "My problem list" screen tries to load ALL keys from preferences and expects string there.
But boolean got mixed in.

Reference: https://trello.com/c/qZ6mp0Ua/172-bug-the-app-crashes-after-right-after-sending-a-report-from-illegal-parking-category


app/src/main/java/lt/vilnius/tvarkau/dagger/module/SharedPreferencesModule.kt
```diff
@@ -17,15 +18,23 @@ class SharedPreferencesModule {
 
     @Provides
     @Singleton
-    internal fun providesSharedPreferences(application: Application): SharedPreferences {
+    @Named(PREFS_NAME)
```
vycius, line 21: Instead of `Named` I would prefer annotation similar to `NewApi`

vycius (approved):
Nice catch

app/src/main/java/lt/vilnius/tvarkau/dagger/module/SharedPreferencesModule.kt
mjurkus: Then there would be key in `Preferences` and annotation.

PR #148, opened 2016-12-29
Drop retrolambda

mjurkus:
Improved overall build speeds by leveraging gradle incremental and kotlin incremental builds.

Ensure that these properties are added to you `gradle.properties`

```
kotlin.incremental=true
org.gradle.daemon=true
```

Although these can be added directly to our project.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,5 +1,4 @@
 plugins {
-    id "me.tatarka.retrolambda" version "3.4.0"
     id 'com.github.ben-manes.versions' version '0.13.0'
 }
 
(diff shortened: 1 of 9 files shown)


vycius:
Have you done any benchmarks on speed gain ?

vycius:
```
kotlin.incremental=true
org.gradle.daemon=true
```
What about adding these lines to projet `gradle.properties` ?

vycius (approved):

mjurkus:
Some numbers:

Cold - Gradle daemon stopped
Warm - with code changes
```
w/ Retrolambda
Cold: 33s
No change 1: 5.2s
No change 2: 1.13s
Warm 1: 5.3s
Warm 2: 4.9s

w/o Retrolambda
Cold: 35s
No change 1: 1.2s
No change 2: 1.05s
Warm 1: 5.1s 
Warm 2: 5.02s
```

Performance increase is marginal, but as more KT code is added they will get only better.

mjurkus:
There is still something that breaks incremental builds as from time to time even with unchanged code full compilation is performed.
Need to look into that later.

Most likely some code generating tool.

PR #151, opened 2017-01-14
Replace birth date with personal code in report submitting

mjurkus:
- replace field
- add personal code validation

image


app/src/main/java/lt/vilnius/tvarkau/utils/PersonalCodeValidator.kt
```diff
@@ -0,0 +1,35 @@
...
+            SECOND_ITERATION_MULTIPLIERS[index] * number
+        }.sum()
+
+        checksum = sum % 11 % 10
+
+        return checksum == control
+    }
+
+    val FIRST_ITERATION_MULTIPLIERS = arrayOf(1, 2, 3, 4, 5, 6, 7, 8, 9, 1)
+    val SECOND_ITERATION_MULTIPLIERS = arrayOf(3, 4, 5, 6, 7, 8, 9, 1, 2, 3)
```
vycius, line 34: IMHO validation is too strict. In this way only people with Lithuanian passports can submit reports. Please discuss with @RenatBuga  about this.

vycius:
Just interested why don't you use vector drawables ?

app/src/main/java/lt/vilnius/tvarkau/entity/Profile.java
```diff
@@ -59,6 +62,10 @@ public void setMobilePhone(String mobilePhone) {
         this.mobilePhone = mobilePhone;
     }
 
+    public String getPersonalCode() {
+        return personalCode;
+    }
+
     public String createJsonData() {
```
vycius: See `equals` method. Or better convert to Kotlin data class to avoid such mistakes

vycius:
@RenatBuga Without any disclaimer where my national identification number will be used, I wouldn't put this in the form.

vycius:
There is some laws and regulations about national identification number usage and security. See https://www.infolex.lt/ta/51494:str7 http://www.rrt.lt/rrt/download/10721/adtai.doc== and search for other laws to see if everything is ok with law. 
IMHO such information should be sent only over HTTPS.

vycius (approved):
From code side looks ok. Fix few small problems and talk about regulations and validation with Renat

app/src/main/java/lt/vilnius/tvarkau/utils/PersonalCodeValidator.kt
RenatBuga: This was discussed with municipality's department for public transport order. The reporter needs to be registered in Lithuania (foreigners can hold LT personal code too without local ID card).

RenatBuga:
@vycius Public disclaimer is present bellow the form

app/src/main/java/lt/vilnius/tvarkau/entity/Profile.java
mjurkus: 👍

PR #171, opened 2017-02-05
Upgrade dependencies

vycius:
Dependencies haven't been updated for quite a long time. 
## Updates to note 
* `ParallelFlowable` https://github.com/ReactiveX/RxJava/releases
* `@BindsInstance` https://github.com/google/dagger/releases

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -68,7 +68,7 @@ ext {
     junitVersion = '4.12'
     supportLibraryVersion = '25.1.1'
     playServicesVersion = '10.0.1'
-    butterKnifeVersion = '8.4.0'
+    butterKnifeVersion = '8.5.1'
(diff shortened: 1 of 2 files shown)


mjurkus (approved):
Generally I'm against to such mass library updates.
Something like - if it's not broken - why fix it.
Such updates can introduce issues we're not aware of.

vycius:
Personally I am also against mass library updates. 
Although sooner or later you will need to upgrade libraries. More time you wait, the higher the probability of breaking something gets due to need to upgrade nearly all libraries. 
Last time libraries had been upgraded  8 Dec 2016
In Trafi we try to update libs every 2-3 weeks with such practices we do upgrades for small amount of libraries, don't have mass library upgrades and catch with recent library changes far more easily.

PR #174, opened 2017-02-11
Implement report type caching and retrieval logic

mjurkus:
Preparing for Map Filter implementation

This will help to avoid hitting API with each filter opening.


app/src/main/java/lt/vilnius/tvarkau/mvp/interactors/ReportTypesInteractorImpl.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * @author Martynas Jurkus
+ */
+class ReportTypesInteractorImpl(
+        val api: LegacyApiService,
+        val ioScheduler: Scheduler
+) : ReportTypesInteractor {
+
+    private val entryCache = mutableListOf<String>()
```
vycius: Why mutablelist over immutable list ? E.g:
```java
    private var entryCache = emptyList<String>()
```

app/src/main/AndroidManifest.xml
```diff
@@ -125,35 +112,35 @@
...
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
-                android:value=".MainActivity"/>
+                android:value=".MainActivity" />
         </activity>
 
         <meta-data
             android:name="io.fabric.ApiKey"
-            android:value="33cce6891f8d5ae5c92235382a89bc0e9d7ca255"/>
+            android:value="33cce6891f8d5ae5c92235382a89bc0e9d7ca255" />
```
vycius: Not related to this PR, but Fabric key can be removed

vycius (approved):
Fantastic

app/src/main/java/lt/vilnius/tvarkau/mvp/interactors/ReportTypesInteractorImpl.kt
mjurkus: well, It's either a mutable variable or mutable list. Not sure which is better.

vycius: If list not modified: items added, removed. I would prefer to have immutable version with mutable variable

mjurkus: Why? Just want to know why `var` is better?

vycius: By using immutable one you are guaranteeing that if something changes new list will be created and modifications will be performed. Also `entryCache::addAll` says that list can be modified and items added more than one time. Which is not true in this case as list is initilized just one time.

PR #178, opened 2017-03-01
Allow report filtering in report list view

mjurkus:
- Remove tab navigation in report list
- Use single toolbar across activities fragments
- Enable filtering in report list view
- Splitting my/all report fragments to separate classes


app/src/main/java/lt/vilnius/tvarkau/fragments/BaseFragment.kt
```diff
@@ -53,7 +57,8 @@ abstract class BaseFragment : Fragment() {
 
     override fun onDestroyView() {
         super.onDestroyView()
-        refWatcher.watch(this)
+        //Something is wrong. Need to discuss this more...
+//        refWatcher.watch(this)
```
mjurkus: I don't understand why `refWatcher.watch(this)` should be attached `onDestroyView()`.
That way it will always spew out reports, that fragment is leaking EventBus, because it's purpose is to keep fragment and receive events even when views are destroyed, but fragment is not.

vycius: Move it to `onDestroy`

vycius:
While clicking got to such state that reports list always crashes with same exception on every open (my reports working correctly). Cleaning app settings solved issue.
giphy 3

```
03-03 22:28:00.250 31596-31596/lt.vilnius.tvarkau E/AndroidRuntime: FATAL EXCEPTION: main
                                                                    Process: lt.vilnius.tvarkau, PID: 31596
                                                                    java.lang.IllegalStateException: Fatal Exception thrown on Scheduler.Worker thread.
                                                                        at rx.android.schedulers.LooperScheduler$ScheduledAction.run(LooperScheduler.java:114)
                                                                        at android.os.Handler.handleCallback(Handler.java:751)
                                                                        at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                        at android.os.Looper.loop(Looper.java:154)
                                                                        at android.app.ActivityThread.main(ActivityThread.java:6119)
                                                                        at java.lang.reflect.Method.invoke(Native Method)
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
                                                                     Caused by: java.lang.IllegalStateException: it must not be null
                                                                        at lt.vilnius.tvarkau.fragments.presenters.AllReportsListPresenterImpl$getReportsForPage$4.call(AllReportsListPresenterImpl.kt:30)
                                                                        at lt.vilnius.tvarkau.fragments.presenters.AllReportsListPresenterImpl$getReportsForPage$4.call(AllReportsListPresenterImpl.kt:11)
                                                                        at rx.Single$11.onSuccess(Single.java:1700)
                                                                        at rx.internal.operators.SingleObserveOn$ObserveOnSingleSubscriber.call(SingleObserveOn.java:88)
                                                                        at rx.android.schedulers.LooperScheduler$ScheduledAction.run(LooperScheduler.java:107)
                                                                        at android.os.Handler.handleCallback(Handler.java:751) 
                                                                        at android.os.Handler.dispatchMessage(Handler.java:95) 
                                                                        at android.os.Looper.loop(Looper.java:154) 
                                                                        at android.app.ActivityThread.main(ActivityThread.java:6119) 
                                                                        at java.lang.reflect.Method.invoke(Native Method) 
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
```

vycius:
Found scenario there this exception always happens (sometimes it happen far quicker):
giphy 4

vycius (approved):
Amazing work. Investigate issue with crash and we are ready to ship this 👍

mjurkus:
>While clicking got to such state that reports list always crashes with same exception on every open (my reports working correctly). Cleaning app settings solved issue.

It looks that you got combination of properties that doesn't have any reports and we always expect at least one report. Good catch!

vycius (approved):

PR #182, opened 2017-03-07
Mixpanel analytics integration

vycius:
Finally new analytics. 
Integration guide https://mixpanel.com/help/reference/android

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -33,6 +33,9 @@ android {
             shrinkResources true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
+        debug {
+            multiDexEnabled = true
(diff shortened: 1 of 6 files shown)


app/build.gradle
```diff
@@ -33,6 +33,9 @@ android {
             shrinkResources true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
+        debug {
+            multiDexEnabled = true
```
mjurkus, line 37: Already need multidex?!

mjurkus (approved):

app/build.gradle
vycius: Yup. This is just for `debug` version. For `prod` version it should be ok.

PR #188, opened 2017-03-23
Fix condition when photo instructions should be displayed

vycius:
Fix condition from https://github.com/vilnius/tvarkau-vilniu/pull/186
Show one time in cycles of 24h, not every 48 hours.


mjurkus (approved):
Ah, good catch.

PR #191, opened 2017-03-24
Make address lookup more error prone

mjurkus:
In some cases `Geocoder` fails to return address.

Fallbacking to latitude/longitude for address if address line is empty.

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/fragments/NewReportFragment.kt
+++ b/app/src/main/java/lt/vilnius/tvarkau/fragments/NewReportFragment.kt
@@ -59,7 +59,6 @@ import pl.aprilapps.easyphotopicker.DefaultCallback
 import pl.aprilapps.easyphotopicker.EasyImage
 import timber.log.Timber
 import java.io.File
-import java.io.IOException
 import java.util.*
 import java.util.Calendar.*
 import javax.inject.Inject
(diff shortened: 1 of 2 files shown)


app/src/main/java/lt/vilnius/tvarkau/fragments/NewReportFragment.kt
```diff
@@ -440,38 +439,51 @@ class NewReportFragment : BaseFragment(),
                     val geocoder = Geocoder(context)
                     val place = PlacePicker.getPlace(context, data)
 
-                    var addresses = listOf<Address>()
+                    val addresses = mutableListOf<Address>()
```
vycius, line 442: I would prefer earlier style. It is much more clear.

app/src/main/java/lt/vilnius/tvarkau/fragments/NewReportFragment.kt
```diff
@@ -440,38 +439,51 @@ class NewReportFragment : BaseFragment(),
...
 
                         try {
-                            addresses = geocoder.getFromLocation(it.latitude, it.longitude, 1)
-                        } catch (e: IOException) {
-                            Timber.e(e)
+                            geocoder.getFromLocation(it.latitude, it.longitude, 1)?.let {
+                                addresses.addAll(it)
+                            }
+                        } catch (e: Throwable) {
+                            Timber.e(GeocoderException(e))
```
vycius, line 452: If exception message doesn't contain coordinates it would be nice to see `latitude` and  `longitude` with exception for easier debugging.

vycius (approved):

app/src/main/java/lt/vilnius/tvarkau/fragments/NewReportFragment.kt
mjurkus: `geocoder.getFromLocation` might return a list of values, a `null` or throw exception.

It's way easier down the line to check if list is empty rather than checking if list is null or empty.

app/src/main/java/lt/vilnius/tvarkau/fragments/NewReportFragment.kt
mjurkus: I guess there will be various IO exceptions as stated in documentation.
How lat/lon would help us in debugging?

vycius: My bad. At this we won't get any valuable information from `latitude`, `longtitude`.

app/src/main/java/lt/vilnius/tvarkau/fragments/NewReportFragment.kt
```diff
@@ -440,38 +439,51 @@ class NewReportFragment : BaseFragment(),
...
                         }
                     }
 
                     if (addresses.isNotEmpty()) {
                         if (addresses.first().locality != null) {
```
vycius: What happen if get got `addresses.first().locality = null` ?

mjurkus: At this point - nothing. How should we approach this? Maybe move this one line up
`addresses.firstOrNull()?.locality != null`?

mjurkus: And then fallback to looking up address from `place`?

vycius: Let's do this :)

PR #201, opened 2017-07-02
Bottom tabs navigation

vycius:
https://trello.com/c/D0427WhF/193-remove-main-screen-add-navigation-drawer
Introduced bottom navigation instead of homescreen. Bottom navigation uses BottomNavigationController. Also added new concept for navigation - NavigationManager with FragmentTransactionExecutor. Looking forward for comments and suggestions.

# TODO
- [x] Map filter doesn't work. Needs investigation.

PR #203, opened 2017-07-12
Fragments navigation icon and screen tracking names

vycius:
@mjurkus I would like your opinion about `BaseFragment` builder.

Fixes regression with fragment navigation introduced in https://github.com/vilnius/tvarkau-vilniu/pull/201

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/fragments/BaseFragment.kt
+++ b/app/src/main/java/lt/vilnius/tvarkau/fragments/BaseFragment.kt
@@ -6,6 +6,7 @@ import android.support.v4.app.Fragment
 import android.support.v7.app.AppCompatActivity
 import com.squareup.leakcanary.RefWatcher
 import lt.vilnius.tvarkau.BaseActivity
+import lt.vilnius.tvarkau.R
 import lt.vilnius.tvarkau.TvarkauApplication
 import lt.vilnius.tvarkau.analytics.Analytics
 import lt.vilnius.tvarkau.backend.LegacyApiService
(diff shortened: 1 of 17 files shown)


app/src/main/java/lt/vilnius/tvarkau/fragments/BaseFragment.kt
```diff
@@ -58,19 +71,42 @@ abstract class BaseFragment : Fragment(), OnBackPressed {
     override fun onResume() {
         super.onResume()
 
-        analytics.trackOpenFragment(activity, javaClass.simpleName)
+        screen?.run {
+            if (titleRes != 0) {
```
mjurkus, line 75: When can it be 0?

app/src/main/java/lt/vilnius/tvarkau/fragments/BaseFragment.kt
```diff
@@ -58,19 +71,42 @@ abstract class BaseFragment : Fragment(), OnBackPressed {
...
+                activity.setTitle(titleRes)
+            }
+
+            trackingScreenName.emptyToNull()?.let {
```
mjurkus, line 79: Here too?

vycius: You have a choice if you want to track fragment or not. If name is empty string tracking is disabled.

app/src/main/java/lt/vilnius/tvarkau/fragments/BaseFragment.kt
vycius: It's annotation default. It means that fragment shouldn't set title.

mjurkus: I don't see any cases where you're using empty title, though.

vycius: For example in NewReport fragment title is set dynamically by report type.

mjurkus (approved):

PR #204, opened 2017-07-21
Don't reload data after onDestroyView

vycius:
Fixes https://console.firebase.google.com/u/1/project/tvarkau-vilniu/monitoring/app/android:lt.vilnius.tvarkau/cluster/2b9f8b8f?duration=2592000000
https://console.firebase.google.com/u/0/project/tvarkau-vilniu/monitoring/app/android:lt.vilnius.tvarkau/cluster/53385b7e?pli=1&reportTypes=REPORT_TYPE_UNSPECIFIED&duration=2592000000

### DIFF
--- a/app/src/main/java/lt/vilnius/tvarkau/fragments/BaseReportListFragment.kt
+++ b/app/src/main/java/lt/vilnius/tvarkau/fragments/BaseReportListFragment.kt
@@ -19,6 +19,7 @@ import lt.vilnius.tvarkau.fragments.views.ReportListView
 import lt.vilnius.tvarkau.rx.RxBus
 import lt.vilnius.tvarkau.views.adapters.ProblemsListAdapter
 import rx.Subscription
+import timber.log.Timber
 
 /**
  * @author Martynas Jurkus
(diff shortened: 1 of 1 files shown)


app/src/main/java/lt/vilnius/tvarkau/fragments/BaseReportListFragment.kt
```diff
@@ -52,16 +53,26 @@ abstract class BaseReportListFragment : BaseFragment(), ReportListView {
                 .subscribeOn(ioScheduler)
                 .observeOn(uiScheduler)
                 .subscribe({
-                    reloadData()
+                    reloadReports = true
```
mjurkus: While the fix would work, I think it's a hack :)

`onError` handler is needed here.

And unsubscribe should be before `super.onDestroyView()`. That should be enough I think.

vycius: This won't work. If unsubscribtion is done in `onDestroyView` ReportsList fragment won't receive an event of applied filters from ReportsFilterFragment.  
We need to keep subscribed `RxBus` after `onDestroyView`.

mjurkus: True.

Still `onError` is needed here.

vycius: True. I will add that.

vycius:
@mjurkus Added on error handler

mjurkus (approved):

PR #218, opened 2018-05-23
Use oauth token for new API

mjurkus:
- Add `TvarkauMiestaApi`
- Implement OAuth token refresh
- Authenticate requests with oauth tokens


vycius (approved):
Assuming that `OauthTokenRefresher` is already tested on another project and working without any deadlocks or other multi-threading issues this PR seems to be very good.

PR #219, opened 2018-05-28
Migrate RxJava to RxJava2

mjurkus:


vycius (approved):
Looks good except Travis is not passing. 
I have no idea what is wrong with Travis. This needs investigation and fixing before merging this PR.

vycius (approved):

PR #228, opened 2018-08-21
Implement all reports list with AAC

mjurkus:
Implementing new API for all reports.

Since API is limited for now - filtering won't work.

All retrieved reports are stored to database first and loaded to UI using PagedList. This ensures that only required items are stored in memory and loaded on demand as needed. This approach also provides out of the box offline state capability (needs to be updated to display network state).

All ACC components are used - https://developer.android.com/jetpack/


app/build.gradle
```diff
@@ -22,6 +22,12 @@ android {
         resConfigs "lt", "en"
 
         vectorDrawables.useSupportLibrary = true
+
+        kapt {
+            arguments {
+                arg("room.schemaLocation", "$projectDir/schemas".toString())
```
vycius: Why do we need `toString`. It should be string already.

app/schemas/lt.vilnius.tvarkau.database.AppDatabase/1.json
```diff
@@ -0,0 +1,157 @@
...
+            "fieldPath": "userId",
+            "columnName": "user_id",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "refNo",
+            "columnName": "ref_no",
+            "affinity": "TEXT",
+            "notNull": false
```
vycius, line 27: How can it be nullable ?

app/schemas/lt.vilnius.tvarkau.database.AppDatabase/1.json
```diff
@@ -0,0 +1,157 @@
...
+          {
+            "fieldPath": "licensePlateNo",
+            "columnName": "license_plate_no",
+            "affinity": "TEXT",
+            "notNull": false
+          },
+          {
+            "fieldPath": "registeredAt",
+            "columnName": "registered_at",
+            "affinity": "TEXT",
```
vycius, line 74: Why are we using TEXT instead of DateTime ?

vycius (approved):
Awesome.
I am not familiar with lifecycle_extensions and room. But from code side it looks good. Just few small questions about DB schema.

app/schemas/lt.vilnius.tvarkau.database.AppDatabase/1.json
mjurkus: Now Avilys and backend works async. So initially `refNo` might be null and later, when document is registered in Avilys record will be updated.

app/schemas/lt.vilnius.tvarkau.database.AppDatabase/1.json
mjurkus: We don't need DateTime - we don't perform any calculations or date manipulations. All we do - display date string. And backend can control how the dates are represented.

PR #229, opened 2018-09-09
Preselect city on app initialization

mjurkus:


vycius (approved):
Looks good.
Check the tests on Travis, because they are failing.

PR #233, opened 2018-10-08
Restore Report Type selection before new report creation

mjurkus:
- Load Report Types from DB
- Pass selected ReportType to new report registration instead of just title.

Report registration is still broken will fix in next PR.


vycius (approved):
Nice

PR #235, opened 2018-10-19
Refurbish Di using Dagger for Android 2.17

mjurkus:
Lots of changes here but business logic is not changed.

This change simplifies dependency injection to the max.
Now fragments and activities don't have to know how dependencies are injected.
When new fragment is added - just do this
```
    @ContributesAndroidInjector
    abstract fun someNewFragment(): SomeNewFragment
```

and that's it.

@vycius


mjurkus:
This also is required for correct `ViewModel` creation at runtime.
Now we lose all `ViewModel`s created  @ `onActivityCreated()` since at the moment only at this point all dependencies are injected.

With this change - all dependencies are ready @ `onCreate()` and `ViewModel`s will be created only once per fragment lifetime.

vycius (commented):
This looks amazing. I haven't knew about `ContributesAndroidInjector`

PR #237, opened 2018-10-30
Retrieve and list current users reports

mjurkus:
Load reports from `ReportsRepository` in `MyReportsListFragment`


vycius:
LGTM. 
P.S Sorry for being late.

PR #238, opened 2018-11-08
Implement authentication via VIISP

mjurkus:
also known as Elektroniniai Valdžios Vartai.

This authentication type will be used when submitting reports about parking violations. User will no longer be required to do that manually.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -108,6 +108,8 @@
                 android:value=".MainActivity" />
         </activity>
 
+        <activity android:name=".activity.ViispLoginActivity" />
+
         <receiver
             android:name="com.mixpanel.android.mpmetrics.InstallReferrerReceiver"
             android:exported="true">
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -108,7 +108,9 @@
                 android:value=".MainActivity" />
         </activity>
 
-        <activity android:name=".activity.ViispLoginActivity" />
+        <activity
+            android:name=".activity.ViispLoginActivity"
+            android:screenOrientation="portrait" />
 
         <receiver
             android:name="com.mixpanel.android.mpmetrics.InstallReferrerReceiver"
--- a/app/src/main/java/lt/vilnius/tvarkau/activity/LoginActivity.kt
+++ b/app/src/main/java/lt/vilnius/tvarkau/activity/LoginActivity.kt
@@ -42,6 +42,7 @@ class LoginActivity : BaseActivity() {
         email_sign_in_button.setOnClickListener { viewModel.attemptLogin() }
         skip_sign_in_button.setOnClickListener { viewModel.signInGuestUser() }
         google_sign_in_button.setOnClickListener { signInWithGoogle() }
+        viisp_sign_in_button.setOnClickListener { signInWithViisp() }
 
         viewModel = withViewModel(viewModelFactory) {
             observeNonNull(errorEvents) {
@@ -66,6 +67,10 @@ class LoginActivity : BaseActivity() {
         finish()
     }
 
+    private fun signInWithViisp() {
+        startActivityForResult(Intent(this, ViispLoginActivity::class.java), RC_VIISP_SIGN_IN)
+    }
+
     private fun signInWithGoogle() {
         val gso = GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
             .requestEmail()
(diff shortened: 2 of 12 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -108,6 +108,8 @@
                 android:value=".MainActivity" />
         </activity>
 
+        <activity android:name=".activity.ViispLoginActivity" />
```
vycius: Possibly, it's better to stick with portrait mode, set label and parent activity as in all other activities.

app/src/main/java/lt/vilnius/tvarkau/activity/ViispLoginActivity.kt
```diff
@@ -0,0 +1,105 @@
...
+    override fun onBackPressed() {
+        if (web_view.canGoBack()) {
+            web_view.goBack()
+        } else {
+            super.onBackPressed()
+        }
+    }
+
+    companion object {
+        private const val VIISP_AUTH_URI = "https://api.tvarkaumiesta.lt/auth/viisp/new"
```
vycius: I don't really hardcoding of API host URL. Host can be injected from `RestAdapterModule`.

app/src/main/java/lt/vilnius/tvarkau/auth/SessionTokenImpl.kt
```diff
@@ -48,9 +48,24 @@ class SessionTokenImpl @Inject constructor(
...
+                .requestAccessToken {
+                    handleAccessToken(it, emitter)
+                }
+        }
+    }
+
     private fun handleAccessToken(response: OAuthResponse, emitter: CompletableEmitter) {
         if (response.isSuccessful) {
-            appPreferences.apiToken.set(gsonSerializer.fromJson(response.body, ApiToken::class.java))
+            appPreferences.apiToken.set(gsonSerializer.fromJson(response.body, ApiToken::class.java), commit = true)
```
vycius: Why ? Commit vs. apply should behave the same.

app/src/main/java/lt/vilnius/tvarkau/activity/LoginActivity.kt
```diff
@@ -84,6 +89,7 @@ class LoginActivity : BaseActivity() {
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
         when (requestCode) {
             RC_SIGN_IN -> viewModel.signInWithGoogle(data)
+            RC_VIISP_SIGN_IN -> viewModel.signInWithViisp(data)
```
vycius, line 92: Don't we care about `resultCode` ? I think we should call this only on `RESULT_OK`.

vycius (approved):
Looks good. Left few minor comments.

app/src/main/java/lt/vilnius/tvarkau/auth/SessionTokenImpl.kt
mjurkus: Nope. Apply is done async. And I had cases where user refresh was done milliseconds before token was set :)

app/src/main/java/lt/vilnius/tvarkau/activity/LoginActivity.kt
mjurkus: Well now we don't set any other results. Have any ideas when we should set other result codes?

app/src/main/java/lt/vilnius/tvarkau/auth/SessionTokenImpl.kt
vycius: Interesting. I always thought that it should behave the same as SharedPreferences instances are singletons.

app/src/main/java/lt/vilnius/tvarkau/activity/LoginActivity.kt
vycius: Doesn't matter. It should be ok as we don't have any other `setResult` inside activity.

app/src/main/java/lt/vilnius/tvarkau/auth/SessionTokenImpl.kt
vycius: Just looked at Android docs https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()
```
As SharedPreferences instances are singletons within a process, it's safe to replace any instance of commit() with apply() if you were already ignoring the return value.

You don't need to worry about Android component lifecycles and their interaction with apply() writing to disk. The framework makes sure in-flight disk writes from apply() complete before switching states.
```

vycius (approved):

